PR #496, opened 2026-03-20
fix: use per-task CARGO_TARGET_DIR to eliminate parallel build lock contention

### LINKED ISSUE #488: [P1] U-11: CARGO_TARGET_DIR contention in parallel task_executor invocations on same project
## Summary

In `task_executor.rs` lines 380-384, `CARGO_TARGET_DIR` is set to `{project_root}/target` for all agent invocations. When multiple tasks run concurrently on the same project (e.g., parallel reviews, retries), they share the same Cargo target directory, causing lock contention and potential build failures.

## Location

`crates/harness-server/src/task_executor.rs` lines 380-384

## Vulnerability Details

```rust
let cargo_env: HashMap<String, String> = [(
    "CARGO_TARGET_DIR".to_string(),
    format!("{}/target", project.display()),
)]
.into();
```

All agent invocations on the same project share the same `CARGO_TARGET_DIR`. This affects:
1. Parallel reviews running `cargo check` simultaneously
2. A review and a fix agent running concurrently
3. Retry tasks that overlap with the original task

## Failure Scenario

1. A periodic review triggers a `cargo check` on project `/home/user/myapp`
2. A user-initiated task also starts a `cargo check` on the same project
3. Both write to `/home/user/myapp/target`
4. Cargo's build lock (`.cargo-lock`) causes one to block or fail
5. Agent times out waiting for the lock, task fails with unhelpful error
6. In worst case (if lock is not respected), build artifacts are corrupted

## Suggested Fix

Use a per-task or per-worktree target directory:

```rust
let task_target = std::env::temp_dir()
    .join("harness-cargo-targets")
    .join(task_id.to_string());

let cargo_env: HashMap<String, String> = [(
    "CARGO_TARGET_DIR".to_string(),
    task_target.display().to_string(),
)]
.into();
```

Clean up the task-specific target directory after the task completes to avoid disk exhaustion.

majiayu000:
## Summary

- All agent invocations previously shared `{project_root}/target` as `CARGO_TARGET_DIR`, causing Cargo's `.cargo-lock` to serialize concurrent tasks on the same project
- Switch to `std::env::temp_dir()/harness-cargo-targets/{task_id}` so each task gets an isolated build directory
- Add `TaskTargetDir` RAII guard that removes the per-task directory when `run_task` returns (success or error), preventing disk exhaustion

Fixes #488

## Test plan

- [ ] `task_target_dir_cleanup_on_drop` — guard removes directory on drop
- [ ] `task_target_dir_no_op_when_absent` — guard does not panic when directory was never created
- [ ] `per_task_target_paths_are_distinct` — different task IDs produce different paths
- [ ] `cargo_target_dir_is_under_temp_not_project_root` — path is under system temp, not project root
- [ ] `cargo test --workspace` — all 481+ tests pass

### DIFF
--- a/crates/harness-server/src/task_executor.rs
+++ b/crates/harness-server/src/task_executor.rs
@@ -353,6 +353,26 @@ fn prepend_constitution(prompt: String, enabled: bool) -> String {
     }
 }
 
+/// RAII guard that removes the task-specific Cargo target directory on drop.
+///
+/// This ensures disk space is reclaimed when `run_task` returns, regardless of
+/// whether it succeeds or fails.
+struct TaskTargetDir(PathBuf);
+
+impl Drop for TaskTargetDir {
+    fn drop(&mut self) {
+        if self.0.exists() {
+            if let Err(e) = std::fs::remove_dir_all(&self.0) {
+                tracing::warn!(
+                    path = %self.0.display(),
+                    error = %e,
+                    "failed to remove task-specific cargo target dir"
+                );
+            }
+        }
+    }
+}
+
 pub(crate) async fn run_task(
     store: &TaskStore,
     task_id: &TaskId,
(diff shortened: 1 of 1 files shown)


majiayu000:
Superseded by #498

PR #678, opened 2026-04-12
fix(q_value_store): log error on malformed experiences_used JSON (RS-10/U-29)

### LINKED ISSUE #675: [P1] RS-10/U-29: Silent JSON parse error in get_experiences_for_task causes incorrect Q-value updates
## Problem

In `crates/harness-server/src/q_value_store.rs:134`, the JSON parse error is silently swallowed:

```rust
let parsed: Vec<String> = serde_json::from_str(&json).unwrap_or_default();
```

If a row's experiences_used column contains malformed JSON, the parse error is replaced with an empty Vec. The caller returns fewer experience IDs than recorded, causing apply_q_update to skip updates for those rules.

## Impact

- Rules used for a task but with unparseable IDs never receive Q-value backprop
- Completely silent: no log, no error — violates U-29 (no silent degradation)
- Storage or serialization bug silently poisons all Q-value updates for affected tasks

## Recommended Fix

```rust
let parsed: Vec<String> = serde_json::from_str(&json).unwrap_or_else(|e| {
    tracing::error!(task_id, row = %json, "q_value_store: failed to parse experiences_used: {e}");
    Vec::new()
});
```

**File:** crates/harness-server/src/q_value_store.rs:134

majiayu000:
## Summary

- Replace `unwrap_or_default()` with `unwrap_or_else(|e| { tracing::error!(...); Vec::new() })` in `get_experiences_for_task`
- Malformed `experiences_used` JSON rows now emit a `tracing::error!` with `task_id` and raw JSON before returning an empty Vec

## Fixes

Closes #675

Violates RS-10 (silent Result discard) and U-29 (no silent degradation) — rules used for a task but with unparseable IDs would silently skip Q-value backprop.

### DIFF
--- a/crates/harness-server/src/q_value_store.rs
+++ b/crates/harness-server/src/q_value_store.rs
@@ -131,7 +131,10 @@ impl QValueStore {
 
         let mut ids = Vec::new();
         for (json,) in rows {
-            let parsed: Vec<String> = serde_json::from_str(&json).unwrap_or_default();
+            let parsed: Vec<String> = serde_json::from_str(&json).unwrap_or_else(|e| {
+                tracing::error!(task_id, row = %json, "q_value_store: failed to parse experiences_used: {e}");
+                Vec::new()
+            });
             ids.extend(parsed);
         }
         ids.sort();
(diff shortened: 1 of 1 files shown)


majiayu000:
Closing in favor of #676 (CI green, same fix)

PR #699, opened 2026-04-12
fix: preserve startup project metadata

majiayu000:
## Summary
- preserve explicit `default_agent` and `max_concurrent` when startup projects are registered into `ProjectRegistry`
- use one startup mapping path for both the auto-registered `default` project and named startup projects
- add regression coverage for startup registration and reopen behavior

## Test plan
- [x] cargo fmt --all
- [x] cargo check
- [x] cargo test --workspace
- [x] RUSTFLAGS="-Dwarnings" cargo check --workspace --all-targets

### DIFF
--- a/crates/harness-server/src/http.rs
+++ b/crates/harness-server/src/http.rs
@@ -14,6 +14,7 @@ use axum::{
     Json, Router,
 };
 use dashmap::DashMap;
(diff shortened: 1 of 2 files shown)


majiayu000:

PR #700, opened 2026-04-12
fix: enforce runtime snapshot schema gate on restore

majiayu000:
## Summary
- reject runtime snapshots whose schema version does not match the current runtime schema
- log startup restore decisions for missing snapshots, schema mismatches, successful restores, and load failures
- add regression coverage for older/newer incompatible snapshots, including reopen behavior

## Test plan
- [x] cargo fmt --all
- [x] cargo check
- [x] cargo test --workspace


majiayu000:
@codex review

PR #702, opened 2026-04-12
fix: align SDK schemas with server payloads

majiayu000:
## Summary
- preserve `approval_request.id` in the TypeScript SDK runtime parser and public type
- align Python timeout events to the shared `timeout_ms` contract used by the TypeScript SDK
- add SDK regression tests that fail on dropped or renamed schema fields

## Test plan
- [x] `cd sdk/typescript && pnpm install && pnpm test`
- [x] `python3 -m unittest sdk.python.tests.test_client`
- [x] `cargo check`
- [x] `cargo fmt --all`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test --workspace`
- [x] `RUSTFLAGS=\"-Dwarnings\" cargo check --workspace --all-targets`


majiayu000:
@codex

PR #705, opened 2026-04-12
fix: align SDK schemas with server responses

majiayu000:
## Summary
- preserve TypeScript approval_request item fields that the server already exposes
- align Python timeout events with the existing timeout_ms payload used by the TypeScript SDK
- add regression coverage for schema drift in both SDKs

## Test plan
- [x] cargo fmt --all
- [x] cargo check
- [x] cargo test --workspace
- [x] python3 -m unittest sdk.python.tests.test_client
- [x] pnpm --dir sdk/typescript test


majiayu000:

Issue #681, opened 2026-04-12
fix: persist task phase and description consistently across DB fallback


majiayu000:
## Summary
Task API fields like `phase` and `description` are not consistently backed by persisted state. After cache eviction or restart, `GET /tasks` and `GET /tasks/{id}` can return defaulted or missing values instead of the real task lifecycle state.

## Impact
- Task observability becomes unreliable after restart or DB fallback
- Cache-hit and cache-miss code paths expose different semantics
- Downstream tooling can make wrong recovery or debugging decisions

## Evidence
- `crates/harness-server/src/task_runner.rs:104-195`
- `crates/harness-server/src/task_db.rs:19,175-231,277-295,747-865`
- `crates/harness-server/src/http.rs:1621-1655`

## Acceptance criteria
- `phase`, `description`, and other externally visible task summary/detail fields have one clear persistence policy
- Restarted tasks and DB-fallback tasks expose the same semantics as in-memory tasks
- Regression tests cover cache-hit vs DB-fallback responses

PR #706, opened 2026-04-12
fix: persist task metadata across DB fallback

majiayu000:
## Summary
- persist task `description` and `phase` in the `tasks` table
- make DB-backed task detail/summary decoding preserve real metadata and fail loudly on invalid phase values
- add regression coverage for migration defaults, cache-vs-DB precedence, and DB fallback semantics

## Test plan
- [x] cargo fmt --all
- [x] cargo test --workspace


majiayu000:
Duplicate of #725 which covers the same fix for #681. Closing to avoid conflict.

PR #711, opened 2026-04-12
refactor: centralize task status semantics

majiayu000:
## Summary
- centralize TaskStatus semantic groups for active, terminal, and interrupted-for-recovery states
- switch recovery, cancellation, intake, and periodic reviewer call sites to shared helpers
- keep Done-only queries and done/failed aggregates separate from terminal semantics

## Test plan
- [x] cargo fmt --all
- [x] cargo check
- [x] cargo test --workspace


majiayu000:

PR #723, opened 2026-04-13
fix(intake): prevent cancelled task re-enqueue and Feishu map leak

majiayu000:
## Summary

Addresses two issues found in review of PR #707:

- **GitHub Issues**: Document and enforce the invariant that cancelled issues stay in `dispatched` so they are not re-enqueued on the next poll cycle. The completion callback now forwards `Cancelled` to `on_task_complete` for consistent lifecycle handling.
- **Feishu**: The `Cancelled` arm in `on_task_complete` returned early before the `dispatched`/`chat_ids` cleanup block, permanently leaking map entries until server restart. Restructured the function so cleanup always runs for Done/Failed/Cancelled; Cancelled is silent (no chat notification) but still removes both map entries.

## Test plan

- [ ] `cargo test --package harness-server --lib intake` — 47 tests pass (includes new `cancelled_task_cleans_up_dispatched_and_chat_ids`)
- [ ] `RUSTFLAGS="-Dwarnings" cargo check --workspace --all-targets` — clean

### DIFF
--- a/crates/harness-server/src/http.rs
+++ b/crates/harness-server/src/http.rs
@@ -1050,6 +1050,12 @@ fn build_completion_callback(
                 task_runner::TaskStatus::Failed => {
                     task.error.as_deref().unwrap_or("unknown error").to_string()
                 }
+                // Forward Cancelled so intake sources can release their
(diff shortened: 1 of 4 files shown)


majiayu000:

PR #724, opened 2026-04-13
fix(db): wrap schema migrations in transactions

### LINKED ISSUE #692: fix: wrap schema migrations in transactions
## Summary
Migrations are currently applied statement by statement, and the schema version is only recorded after all statements succeed. If a multi-statement migration fails midway, the database can be left in a partially applied state.

## Impact
- Schema and `schema_migrations` can diverge
- Restarts can retry against a half-mutated schema
- Recovery from migration failures becomes harder to reason about

## Evidence
- `crates/harness-core/src/db.rs:195-221`

## Acceptance criteria
- Each migration version executes inside its own transaction when possible
- Partial migration application does not leave version tracking inconsistent
- Errors make it obvious where the migration stopped

majiayu000:
Closes #692

## Summary

- Each migration version now executes inside its own transaction (all SQL statements + the `schema_migrations` INSERT are committed atomically)
- On failure, the transaction is rolled back automatically — schema and version tracking stay consistent
- Error messages include `(transaction rolled back)` so the failure point is immediately obvious

## Test plan

- [ ] Existing tests pass: `migrator_applies_pending_migrations`, `migrator_is_idempotent_on_rerun`, `migrator_tolerates_duplicate_column_on_alter_table`
- [ ] New test `migrator_rolls_back_partial_migration_on_failure` verifies that a multi-statement migration with a bad second statement leaves no schema change and no version record

### DIFF
--- a/crates/harness-core/src/db.rs
+++ b/crates/harness-core/src/db.rs
@@ -193,12 +193,13 @@ impl<'a> Migrator<'a> {
         pending.sort_by_key(|m| m.version);
 
         for migration in pending {
+            let mut tx = self.pool.begin().await?;
             for stmt in migration.sql.split(';') {
                 let stmt = stmt.trim();
                 if stmt.is_empty() {
                     continue;
                 }
-                if let Err(e) = sqlx::query(stmt).execute(self.pool).await {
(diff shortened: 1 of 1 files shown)


majiayu000:
Duplicate of #697 which already passes CI. Closing to avoid conflict.

PR #732, opened 2026-04-13
fix(prompts): remove stale worktree dir before git worktree add

majiayu000:
## Summary
- `git worktree prune` only removes *registered* worktrees whose directories are gone; it cannot remove a stale directory left by a prior task that was SIGKILL'd after `mkdir` but before registration
- Add `rm -rf /tmp/harness-pr-{pr_number} 2>/dev/null || true` before `git worktree add` in `continue_existing_pr()` to handle this case
- Apply the same fix for `/tmp/harness-review-{pr}` in `review_prompt()`

## Test plan
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` — no warnings
- [ ] `cargo test --workspace` — all tests pass


majiayu000:
Duplicate of #735. Closing.

PR #735, opened 2026-04-13
fix(prompts): remove stale worktree directory before git worktree add

### LINKED ISSUE #730: [P1] RS-13: git worktree prune cannot remove stale directories — PR review/fix tasks fail without recovery path
## Description

After the fix in #722, PR review and fix agents are instructed to create a git worktree at a deterministic path before making any changes:

```
git fetch origin {branch}
git worktree prune
git worktree add /tmp/harness-pr-{pr_number} {branch}
```

**The problem**: `git worktree prune` only removes *registered* worktrees whose directories no longer exist. It does **not** remove a directory that exists but was never (or is no longer) registered as a git worktree — for example, one left behind when a prior task was SIGKILL'd after `mkdir` but before `git worktree add` completed its registration.

When this stale directory exists, `git worktree add` fails with:
```
fatal: '/tmp/harness-pr-<N>' already exists
```

The agent has no recovery instruction for this case. Since the prompt also explicitly prohibits `git checkout` in the main repo, the agent has no valid fallback and the task fails without producing useful output.

**Same issue applies** to the review prompt path that uses `/tmp/harness-review-{pr}`.

## Affected Files

- `crates/harness-core/src/prompts.rs:57-58` (continue\_existing\_pr prompt)
- `crates/harness-core/src/prompts.rs:364-365` (review\_prompt path)

## Recommended Fix

Add a cleanup step before `git worktree add` to handle both cases (registered stale entry and unregistered leftover directory):

```
git fetch origin {branch}
git worktree prune
rm -rf /tmp/harness-pr-{pr_number} 2>/dev/null || true
git worktree add /tmp/harness-pr-{pr_number} {branch}
```

This removes the directory if it exists regardless of registration status, so the subsequent `git worktree add` always succeeds on a clean slate.

majiayu000:
## Summary

- Adds `rm -rf /tmp/harness-pr-{pr_number} 2>/dev/null || true` before `git worktree add` in `continue_existing_pr`
- Adds `rm -rf /tmp/harness-review-{pr} 2>/dev/null || true` before `git worktree add` in `review_prompt`

`git worktree prune` only removes registered worktrees whose backing directories are gone. It does **not** handle directories that exist but were never registered as worktrees (e.g. left by a SIGKILL'd task that completed `mkdir` but was killed before `git worktree add` finished registration). This caused `git worktree add` to fail with `fatal: '/tmp/harness-pr-<N>' already exists`, leaving the agent with no recovery path.

The `rm -rf ... 2>/dev/null || true` step removes any leftover directory unconditionally before the `git worktree add`, ensuring the command always starts clean regardless of whether the path is a stale registered worktree or an unregistered leftover.

Fixes #730

## Test plan
- [x] `cargo fmt --all` — no changes
- [x] `cargo test --workspace` — all tests pass

### DIFF
--- a/crates/harness-agents/src/claude.rs
+++ b/crates/harness-agents/src/claude.rs
@@ -479,14 +479,8 @@ mod tests {
         let req_no_phase = AgentRequest::default();
 
         assert_eq!(agent.resolve_model(&req_planning), "claude-opus-4-6");
-        assert_eq!(
-            agent.resolve_model(&req_execution),
-            "claude-sonnet-4-6"
-        );
-        assert_eq!(
-            agent.resolve_model(&req_validation),
-            "claude-opus-4-6"
-        );
+        assert_eq!(agent.resolve_model(&req_execution), "claude-sonnet-4-6");
+        assert_eq!(agent.resolve_model(&req_validation), "claude-opus-4-6");
         // No phase → falls back to default_model
         assert_eq!(agent.resolve_model(&req_no_phase), "default-model");
     }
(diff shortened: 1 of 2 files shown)


majiayu000:

PR #729, opened 2026-04-13
refactor: unify project identity across registry, queue, and DB (#684)

majiayu000:
## Summary

- Add `ProjectRegistry::resolve_id_for_path` to reverse-look up a canonical path back to its semantic registry ID (e.g. `"harness"`)
- Introduce a single choke point in `enqueue_task` / `enqueue_task_background` that resolves the canonical path to a semantic `project_id`; queue limits, dedup, and DB storage all use this uniform key
- Add `project_id: String` field to `TaskState` with `#[serde(default)]` so legacy deserialized rows (empty string) round-trip cleanly
- DB migration v13: `ALTER TABLE tasks ADD COLUMN project_id TEXT`; INSERT writes semantic ID; dedup query matches on `project_id` with OR-fallback to `project` path column for legacy NULL rows
- Fix all `TaskState` literal constructors in tests to include `project_id`

Closes #684

## Test plan

- [x] `cargo fmt --all` — clean
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — no warnings
- [x] `cargo test --workspace` — all pass
- [ ] Manual: submit task to registered project, verify `project_id` column in `tasks.db` equals semantic ID
- [ ] Manual: submit duplicate issue to same project, verify dedup returns existing task ID

PR #712, opened 2026-04-12
fix: canonicalize project identity by registry id

majiayu000:
## Summary
- use registry project ids as the canonical internal identity for queue admission and config resolution
- keep raw project paths as HTTP boundary aliases and migrate legacy per-root concurrency config deterministically
- update canonicalization-sensitive tests to use real directories

## Test plan
- [x] cargo test -p harness-server default_project_service_ -- --nocapture
- [x] cargo fmt --all
- [x] cargo clippy --workspace --all-targets -- -D warnings
- [x] cargo test --workspace
- [x] RUSTFLAGS=\"-Dwarnings\" cargo check --workspace --all-targets

Issue #684, opened 2026-04-12
refactor: unify project identity across registry, queue, and config resolution


majiayu000:
## Summary
The system currently uses multiple project identities at once: project id, canonical project path, and project-root-based config resolution. These are not consistently unified across runtime subsystems.

## Impact
- Project-level queue limits, default agent selection, and active flags can diverge
- The same logical project is keyed differently across subsystems
- Future project-scoped policy work remains brittle until the identity model is unified

## Evidence
- `crates/harness-server/src/project_registry.rs:7-18`
- `crates/harness-server/src/http/task_routes.rs:17-109`
- `crates/harness-workflow/src/task_queue.rs:421-446`
- `crates/harness-core/src/config/resolve.rs:31-87`

## Acceptance criteria
- Internal project identity has a single canonical form
- Registry, queue, and config resolution all operate on that same identity
- Project id is either a stable alias or removed from internal decision paths

PR #736, opened 2026-04-13
fix(task): prevent duplicate PRs for terminal tasks with pr_url

### LINKED ISSUE #731: fix: dedup misses open PRs from terminal tasks — causes duplicate PRs per issue
## Problem

Task dedup (`find_active_duplicate`) only checks for **active** (non-terminal) tasks with the same `project + external_id`. When a previous task has reached terminal state (done/failed) but its PR is still open on GitHub, a new batch submission creates a second task and a second PR for the same issue.

**Example**: Issue #684 got two PRs:
- #712 (`feat/issue-684-canonical-project-id`) — from first batch, task went terminal
- #729 (`feat/project-identity-unification-684`) — from second batch, dedup didn't catch it

## Root Cause

`check_duplicate()` in `task_routes.rs:70-83` → `find_active_duplicate()` in `task_runner.rs:740-778` only matches tasks in active statuses (`pending`, `implementing`, etc.). Terminal tasks with open PRs are invisible to dedup.

## Proposed Fix

Add a second dedup layer: when the DB finds a **terminal** task for the same `project + external_id` that has a non-null `pr_url`, check whether that PR is still open. If open, skip task creation and return the existing task ID.

Alternatively, during triage phase, instruct the agent to check `gh pr list --search "issue:N"` and skip if an open PR already targets the issue.

## Files

- `crates/harness-server/src/http/task_routes.rs` — `check_duplicate()`
- `crates/harness-server/src/task_runner.rs` — `find_active_duplicate()`
- `crates/harness-server/src/task_db.rs` — `find_active_duplicate()` SQL query

majiayu000:
## Summary

- Adds a second dedup layer in task submission: when a `done` task already has a `pr_url` for the same `project + external_id`, re-submissions return the existing task ID instead of spawning a new agent that creates a duplicate PR.
- `find_terminal_with_pr()` in `task_db.rs` queries only `done` tasks (not `failed`/`cancelled`) so retries after failure are always permitted.
- `find_terminal_pr_duplicate()` in `task_runner.rs` is cache-first with DB fallback; fail-open on DB errors.
- `check_pr_duplicate()` in `task_routes.rs` is called in both `enqueue_task` and `enqueue_task_background`, before permit acquisition, matching the existing active-dedup pattern.

Fixes #731. Root cause of issue #684 receiving two PRs (#712 and #729).

## Test plan

- [ ] `find_terminal_with_pr_returns_done_task_with_pr_url` — done task with pr_url → match
- [ ] `find_terminal_with_pr_ignores_done_task_without_pr_url` — done task, pr_url NULL → no match
- [ ] `find_terminal_with_pr_ignores_failed_task` — failed task with pr_url → no match
- [ ] `find_terminal_with_pr_ignores_cancelled_task` — cancelled task with pr_url → no match
- [ ] All existing `find_active_duplicate_*` tests still pass
- [ ] `cargo test --package harness-server` green

### DIFF
--- a/crates/harness-server/src/http/task_routes.rs
+++ b/crates/harness-server/src/http/task_routes.rs
@@ -86,13 +86,13 @@ async fn check_duplicate(
 ///
 /// This is the second dedup layer — it catches cases where the first batch completed
 /// (task is Done + pr_url set) but a re-submission would create a duplicate PR.
-async fn check_pr_duplicate(
+fn check_pr_duplicate(
     tasks: &Arc<crate::task_runner::TaskStore>,
     project_id: &str,
     req: &task_runner::CreateTaskRequest,
 ) -> Option<task_runner::TaskId> {
     let ext_id = req.external_id.as_deref()?;
-    let (existing_id, pr_url) = tasks.find_terminal_pr_duplicate(project_id, ext_id).await?;
+    let (existing_id, pr_url) = tasks.find_terminal_pr_duplicate(project_id, ext_id)?;
     tracing::info!(
         existing_task = %existing_id.0,
         external_id = %ext_id,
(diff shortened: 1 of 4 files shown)


majiayu000:

Issue #765, opened 2026-04-14
[P1] U-17: startup Source-A redispatch silently leaves task pending on acquire() failure


majiayu000:
## Description

In `http.rs` the startup PR-based redispatch (Source A, lines ~796–803) silently returns
when `task_queue.acquire()` fails (memory pressure or queue full), leaving the recovered
task stuck in `pending` forever — or until the next server restart.

```rust
// http.rs ~796
Err(e) => {
    tracing::error!(...);
    return; // ← task stays pending, no failure status written
}
```

Source B (checkpoint-based redispatch, lines ~990–1016) was introduced in the same PR and
handles the same failure correctly: it calls `mutate_and_persist(Failed)` and fires the
completion callback.

## Root Cause

Source A was written before Source B; its error path was not updated to match the
fail-fast pattern introduced in Source B.

## Impact

- `task_queue.acquire()` can fail immediately when:
  1. `memory_pressure` flag is set (configured threshold breached)
  2. `max_queue_size` is reached at startup (unusual but possible with low caps)
- When either condition holds at startup, Source A tasks silently remain in `pending`
  with `pr_url` set. On every subsequent restart the same failure repeats.
- Operators cannot distinguish a "stuck pending" task from a normal queued task.
- The completion callback is never fired, so intake sources (e.g. GitHub Issues poller)
  never remove the task from their `dispatched` map — causing a permanent dedup deadlock
  for that issue.

## Recommended Action

Apply the same error-handling pattern from Source B to Source A at `http.rs:~796–803`:

```rust
Err(e) => {
    let reason = format!("startup recovery: failed to acquire concurrency permit: {e}");
    tracing::error!(task_id = ?task.id, "{reason}");
    if let Err(pe) = task_runner::mutate_and_persist(
        &state.core.tasks,
        &task.id,
        move |s| {
            s.status = task_runner::TaskStatus::Failed;
            s.error = Some(reason);
        },
    )
    .await
    {
        tracing::error!(task_id = ?task.id,
            "startup recovery: failed to persist failed status: {pe}");
    }
    if let Some(cb) = &state.intake.completion_callback {
        if let Some(final_state) = state.core.tasks.get(&task.id) {
            cb(final_state).await;
        }
    }
    return;
}
```

## File Reference

`crates/harness-server/src/http.rs:796–803`

PR #766, opened 2026-04-14
fix(server): apply fail-fast error handling to startup Source-A acquire failure

majiayu000:
## Summary
- When `task_queue.acquire()` fails during startup PR-based redispatch (Source A, `http.rs:796`), the previous code only logged the error and returned — leaving the task pending indefinitely and causing a permanent dedup deadlock in intake sources.
- Applied the same fail-fast pattern already used in Source B (lines ~990-1016): call `mutate_and_persist(Failed)` with a descriptive error message, then fire the completion callback before returning.
- Fixes U-17 (error handling completeness) violation identified in code review.

## Test plan
- [x] `cargo fmt --all` — clean
- [x] `cargo test --workspace` — all tests pass


majiayu000:
Duplicate of #767 (same fix for #765). Closing older one.

Issue #791, opened 2026-04-17
fix(dedup): terminal-PR dedup blocks resubmission after PR is closed


majiayu000:
## Problem

`find_terminal_with_pr` in `crates/harness-server/src/task_db.rs` (line 338) treats any `done` task with a non-null `pr_url` as a valid duplicate, regardless of whether the PR is still open. When a PR is later **closed without merging** (e.g. after a deep conflict that triggered the close-and-redo strategy), the original task record still has `status='done'` and `pr_url` set — so any resubmission of the same `external_id` is silently blocked and returns the stale task.

## Repro

1. Submit issue N → harness creates task T1 that opens PR P1 → T1 goes to `done`.
2. P1 later gets closed (deep-conflict close, CI break, whatever).
3. Resubmit issue N via `POST /tasks/batch` with `issues: [N]`.
4. **Expected:** new task is created to retry the issue on latest main.
5. **Actual:** dedup returns T1's ID (already done, pointing to closed P1). Nothing runs.

This was hit today when trying to resubmit #625 after #786 was closed-for-redo — the prompt workaround was needed to bypass dedup.

## Current SQL

```sql
SELECT id, pr_url FROM tasks
WHERE project = ? AND external_id = ? AND status = 'done' AND pr_url IS NOT NULL
ORDER BY created_at DESC LIMIT 1
```

## Fix options

1. **Cache PR state in the task record** — add `pr_state` column (open/merged/closed) updated when the orchestrator observes state transitions (e.g. after merge/close webhooks or on rebase turn). Dedup requires `pr_state IN ('open','merged')`.
2. **Lazy probe at dedup time** — when `find_terminal_with_pr` returns a candidate, call `gh pr view <pr>` to check state. Cheap, but violates the "no gh in harness crates" rule unless delegated to an agent.
3. **Clear `pr_url` on close** — when the orchestrator detects a PR transitions to closed-without-merge, NULL out `pr_url` on the originating task so it stops matching dedup. Keeps the historical record via a separate `closed_pr_url` column.

Option 1 is the most robust — it fits the existing schema-first pattern in `task_db.rs` and doesn't add external calls to the dedup hot path.

## Acceptance

- Resubmitting an issue whose previous PR was closed (unmerged) creates a new task.
- Resubmitting an issue whose previous PR was merged still dedups (no regression).
- Resubmitting while a PR is open still dedups via the Layer 1 active-task check (no regression).

majiayu000:
## Context from closed PR #793

[...]

PR #796, opened 2026-04-17
feat(scheduler): periodic retry loop for stalled open issues (#794)

### LINKED ISSUE #794: feat(scheduler): auto-retry loop for open issues with no active work
## Problem

Harness has **two** entry points that create tasks:

1. **GitHub webhook** — fires only on new issue events (open/label/comment). If a webhook-sourced task fails, the issue is never re-picked-up unless a human edits the issue to refire a webhook.
2. **Manual `POST /tasks(/batch)`** — requires a human or external script to resubmit.

Neither handles the **long tail of stalled open issues**: issues that are still open, have no active PR, and whose last task either failed or had its PR closed. They silently rot until a human notices and manually resubmits.

Observed today: 12+ open issues (including #625, #640, #684, #695, #717, #747–#754) that had previously been submitted and failed. None would ever be retried without manual intervention. Compounded by the dedup gap in #791 which actively blocks simple resubmission.

## Requested behavior

Add a periodic **retry scheduler** (similar in shape to `periodic_reviewer`) that:

1. Every N minutes (configurable, default 30min), lists open issues in the configured repo(s) via `gh issue list --state open --json number,updatedAt,labels`.
2. For each issue, checks the local `tasks` table:
   - If there is an **active task** (status in `pending/running/waiting`) for that `external_id` → skip.
   - If there is an **open PR** linked to a task for that `external_id` → skip.
   - If the most recent task is `failed` AND ≥ cooldown-period old (default 6 hours), enqueue a fresh task.
   - If there is no task at all → enqueue.
3. Respects a per-issue **retry cap** (default 3 retries over 72h) — after that, label the issue `harness:stuck` and stop trying so it surfaces to a human.
4. Emits a summary event at each tick: `retries_enqueued`, `skipped_active`, `skipped_capped`, so it is observable.

## Interplay with #791

This scheduler would **also** need the dedup gap in #791 to be fixed — otherwise `find_terminal_pr_duplicate` would re-block the retry for any issue whose previous PR was closed unmerged. Implement #791 first, then this.

## Interplay with `periodic_reviewer`

`periodic_reviewer` scans the **codebase** for new violations. This new scheduler scans **existing issues** for stalled work. They are orthogonal — one finds new work, the other keeps old work moving.

## Acceptance

- With the scheduler enabled and no manual submissions, an open issue whose PR was closed unmerged is automatically retried on the next tick (after cooldown).
- After 3 failed retries for the same issue, further retries are suppressed and the `harness:stuck` label is applied.
- Active/in-flight tasks are never duplicated by the retry loop.
- The scheduler can be disabled via config for users who prefer fully-manual intake.

## Priority

p2 — unblocks a large class of stuck issues once #791 lands. Not urgent today, but the permanent fix for the symptom "many open issues, nothing is happening".

majiayu000:
## Summary

- Adds `RetrySchedulerConfig` (`enabled`, `interval_secs`, `stale_threshold_mins`, `cooldown_mins`, `max_retries`) to `HarnessConfig` under `[retry_scheduler]`; disabled by default — zero behaviour change for existing deployments
- New `TaskDb::list_stalled_tasks()` / `TaskStore::list_stalled_tasks()` query: selects active-status tasks with a non-null `external_id` that haven't been updated in over `stale_threshold_mins` minutes
- New `crates/harness-server/src/periodic_retry.rs`: background loop that detects stalled tasks, applies cooldown and retry-cap logic, re-enqueues via `task_routes::enqueue_task()`, and emits `periodic_retry:attempt:{id}` / `periodic_retry:stuck:{id}` / `periodic_retry:summary` events
- `EventStore::persist_retry_summary()` convenience method; `periodic_retry:*` hooks added to purge exclusion list alongside `periodic_review:*` so attempt history survives retention purges
- `Scheduler::start()` wires in `crate::periodic_retry::start()`

Closes #794. Depends on #791 for the `pr_state` guard (excluded from query until that column lands).

## Test plan

- [ ] 9 unit/integration tests in `periodic_retry.rs` — stall detection, terminal exclusion, null-external-id skip, project filter, summary event shape and decision, disabled-flag path
- [ ] All 1 400+ existing tests pass (`cargo test --workspace`)
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` clean
- [ ] `cargo fmt --all` applied
- [ ] Pre-commit hook (fmt + clippy + test) passed

## Notes

- Direct `gh` / `git` calls inside harness crates are forbidden (CLAUDE.md); the stuck-label path enqueues an agent task with a prompt to apply the label instead
- No `Cargo.toml` version bump per project convention (version-bumps happen at release time only)

Signed-off-by: majiayu000 <1835304752@qq.com>

### DIFF
--- a/crates/harness-core/src/config.rs
+++ b/crates/harness-core/src/config.rs
@@ -49,6 +49,8 @@ pub struct HarnessConfig {
     pub intake: IntakeConfig,
     #[serde(default)]
     pub review: ReviewConfig,
+    #[serde(default)]
+    pub retry_scheduler: RetrySchedulerConfig,
     /// Projects declared in the config file. Registered on server startup.
     #[serde(default)]
     pub projects: Vec<ProjectEntry>,
@@ -520,6 +522,40 @@ mod tests {
     fn harness_config_includes_review() {
         let config = HarnessConfig::default();
         assert!(!config.review.enabled);
+    }
+
+    #[test]
+    fn retry_scheduler_config_defaults() {
+        let config = RetrySchedulerConfig::default();
+        assert!(!config.enabled);
+        assert_eq!(config.interval_secs, 300);
+        assert_eq!(config.stale_threshold_mins, 60);
+        assert_eq!(config.cooldown_mins, 120);
+        assert_eq!(config.max_retries, 3);
+    }
+
+    #[test]
+    fn retry_scheduler_config_deserializes_from_toml() {
+        let toml_str = r#"
+            enabled = true
+            interval_secs = 600
+            stale_threshold_mins = 30
+            cooldown_mins = 60
+            max_retries = 5
+        "#;
+        let config: RetrySchedulerConfig = toml::from_str(toml_str).unwrap();
+        assert!(config.enabled);
+        assert_eq!(config.interval_secs, 600);
+        assert_eq!(config.stale_threshold_mins, 30);
+        assert_eq!(config.cooldown_mins, 60);
+        assert_eq!(config.max_retries, 5);
+    }
+
+    #[test]
+    fn harness_config_includes_retry_scheduler() {
+        let config = HarnessConfig::default();
+        assert!(!config.retry_scheduler.enabled);
+        assert_eq!(config.retry_scheduler.interval_secs, 300);
     }
 
     #[test]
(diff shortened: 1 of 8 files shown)


majiayu000:
@codex review

Issue #806, opened 2026-04-18
[P1] LOGIC: gc_agent.run() lacks timeout in task-completion critical path


majiayu000:
## Description

`QualityTrigger::check_and_maybe_trigger` is awaited directly in the task-completion handler (`http.rs:430`) — not spawned as a background task. Inside that function, `gc_agent.run()` is awaited with no timeout:

```rust
// quality_trigger.rs ~line 283
match self
    .gc_agent
    .run(&project, &events, &[], agent.as_ref())
    .await
{
    Ok(report) => { ... }
    Err(e) => { ... }
}
```

The cross-review path *does* have a 120-second timeout (line ~220), but the GC run that follows has none. If the spawned agent becomes unresponsive (e.g., network hang, out-of-budget stall), the task-completion handler blocks indefinitely. This prevents:

1. The task SSE stream from emitting the terminal event.
2. Any subsequent tasks queued behind the completed one from being admitted.

## Root Cause

Added in `fix(learn): repair broken feedback loop` (commit 79226b2): `auto_adopt_matching` is called after `gc_agent.run()`, but no outer `tokio::time::timeout` guards the run.

## File Reference

- `crates/harness-server/src/quality_trigger.rs` — lines 268–295 (GC run block)
- `crates/harness-server/src/http.rs:430` — call site (direct `.await`, not `tokio::spawn`)

## Recommended Action

Wrap the `gc_agent.run(...)` call in `tokio::time::timeout` (e.g., matching `stall_timeout_secs` from `ConcurrencyConfig`, or a dedicated `gc_run_timeout_secs` config field). If the timeout fires, log a warning and return without blocking the caller.

```rust
match tokio::time::timeout(
    Duration::from_secs(self.gc_run_timeout_secs),
    self.gc_agent.run(&project, &events, &[], agent.as_ref()),
)
.await
{
    Ok(Ok(report)) => { /* auto-adopt */ }
    Ok(Err(e)) => tracing::warn!("quality_trigger: gc_agent.run failed: {e}"),
    Err(_) => tracing::warn!("quality_trigger: gc_agent.run timed out"),
}
```

PR #808, opened 2026-04-18
fix(quality_trigger): add timeout to gc_agent.run() in auto-GC path

majiayu000:
## Summary

- `gc_agent.run()` was awaited without a timeout inside `check_and_maybe_trigger`, which is called from the task SSE handler. A hung agent (network stall, over-budget) would block the completion handler indefinitely, stalling the SSE stream and starving queued tasks.
- Added `gc_run_timeout_secs` field to `GcConfig` (default 120 s, matching the existing cross-review timeout).
- Wrapped `gc_agent.run()` with `tokio::time::timeout`; logs a warning and returns on expiry without blocking the caller.

## Test plan

- [ ] `cargo check --workspace --all-targets` passes
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [ ] `cargo test --package harness-server --package harness-core` passes
- [ ] Config field `gc_run_timeout_secs` is deserializable from TOML (serde default = 120)

### DIFF
--- a/crates/harness-core/src/config/misc.rs
+++ b/crates/harness-core/src/config/misc.rs
@@ -204,6 +204,9 @@ pub struct GcConfig {
     /// Tools allowed during GC agent execution. Default: ["Read", "Grep", "Glob"].
     #[serde(default)]
     pub allowed_tools: Option<Vec<String>>,
+    /// Seconds before an auto-triggered gc_agent.run() call is cancelled. Default: 120.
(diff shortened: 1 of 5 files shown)


majiayu000:
Superseded by #811 which takes the same approach (wrap gc_agent.run in tokio::time::timeout for #806). Closing as duplicate.

Issue #807, opened 2026-04-18
[P1] LOGIC: non-exhaustive AutoAdoptPolicy guard in auto_adopt_matching silently no-ops future variants


majiayu000:
## Description

In `auto_adopt_matching` (gc_agent.rs), after the early-return for `AutoAdoptPolicy::Off`, the loop contains a second policy check that is unreachable with current variants but silently no-ops for any future variant:

```rust
// gc_agent.rs:316-332
if matches!(policy, AutoAdoptPolicy::Off) || draft_ids.is_empty() {
    return Vec::new();
}
let mut adopted = Vec::new();
for id in draft_ids {
    // ...
    if !matches!(policy, AutoAdoptPolicy::RulesOnly) {
        // ↑ Dead code for current variants — only RulesOnly can reach here
        // ↑ A future variant (e.g. AllDrafts) would silently skip every draft
        continue;
    }
    // ...
}
```

Since `Off` exits early, only `RulesOnly` reaches the loop. The inner `if !matches!(policy, AutoAdoptPolicy::RulesOnly)` is permanently false today, but if a new variant (e.g. `AllDrafts`) is added to `AutoAdoptPolicy`, it silently hits `continue` for every draft — behaving identically to `Off` with no error or log. This violates **U-23 (禁止静默降级)**: unsupported policies must fail explicitly, not degrade silently.

## File Reference

- `crates/harness-gc/src/gc_agent.rs:316-354` (`auto_adopt_matching`)
- `crates/harness-core/src/config/misc.rs` — `AutoAdoptPolicy` enum definition

## Recommended Action

Replace the inner check with a `match` that is exhaustive (and will fail to compile when new variants are added):

```rust
match policy {
    AutoAdoptPolicy::Off => return Vec::new(),  // unreachable here; move to early return
    AutoAdoptPolicy::RulesOnly => {
        if draft.signal.remediation != RemediationType::Rule { continue; }
        // ... prefix check ...
    }
}
```

Or alternatively, add a `compile_error!`-style exhaustiveness check via `#[non_exhaustive]` + `let AutoAdoptPolicy::RulesOnly = policy else { continue; }` with an explicit `unreachable!()` for the `Off` arm.

PR #809, opened 2026-04-18
fix(gc): replace auto_pr bool with exhaustive AutoAdoptPolicy enum (U-23)

majiayu000:
## Summary
- Introduce `AutoAdoptPolicy { Off, RulesOnly }` enum in `harness-core::config::misc`, replacing the bare `auto_pr: bool` field in `GcConfig`
- Replace the boolean check in `handlers/gc.rs` with an exhaustive `match` so any future variant causes a compile error instead of a silent no-op (U-23)
- Update test helpers and assertions to use the new enum

## Test plan
- [ ] `RUSTFLAGS="-Dwarnings" cargo check --workspace --all-targets` — clean
- [ ] `cargo test --workspace` — all pass

### DIFF
--- a/crates/harness-core/src/config/misc.rs
+++ b/crates/harness-core/src/config/misc.rs
@@ -176,6 +176,22 @@ impl Default for ValidationConfig {
     }
 }
 
+/// Controls whether and how `gc adopt` automatically dispatches an agent task after writing
+/// draft artifacts to disk.
+///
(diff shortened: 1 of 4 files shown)


majiayu000:
Superseded by #812 which implements the exhaustive AutoAdoptPolicy match for #807. Closing as duplicate.

PR #816, opened 2026-04-18
feat(server): system overview page at /overview with live aggregated API

majiayu000:
## Summary

- Adds a new system-level overview page at `GET /overview` (before any single project is selected), backed by `GET /api/overview` that aggregates live data across all projects, runtimes, and the event store.
- Ports the Claude Design export into the server: inline CSS/JS prototype is embedded via `include_str!`, the mocked JS data is swapped for a 5-second fetch loop that drives KPIs, stacked-area fleet throughput, project table, runtime cards, cluster heatmap, live activity feed, and heuristic alerts.
- Metrics that the project doesn't yet track (per-agent tokens, runtime CPU / RAM) return `null` and the UI falls back to a muted "not wired" note instead of fabricating numbers.

## Changes

- `task_db.rs` / `task_runner.rs` / `services/task.rs` — new `count_done_since` and `done_per_project_hour_since` queries (reuse existing `idx_tasks_status_updated` and `idx_tasks_project_status_updated` indexes).
- `handlers/overview.rs` — new `GET /api/overview` handler, reuses `TaskService`, `RuntimeHostManager`, `RuntimeProjectCache`, and the quality grader / event store.
- `overview.rs` + `static/overview.html` — page handler + design file with live fetch; auth middleware exempts `/overview` for the same reason `/` is exempt (no embedded token, public endpoint only).
- `http.rs` — routes `/overview` and `/api/overview`.

## Test plan

- [x] `RUSTFLAGS="-Dwarnings" cargo check --workspace --all-targets`
- [x] `cargo fmt --all -- --check`
- [x] `cargo test -p harness-server --lib` (776/776 passed)
- [x] New handler covered by 4 tests, including end-to-end JSON shape assertion
- [ ] Manual: start server, open `http://localhost:9800/overview`, verify KPI band / projects table / runtimes / feed render and refresh every 5 s
- [ ] Manual: confirm palette FAB persists choice across reload and shares `harness.palette` key with `/`

## Design

Ported from the Claude Design handoff bundle (`xw4fxxcTtCUBzKbpJkQueA`, `overview.html`). The chat transcript established a "system overview before picking a project" scope; this PR builds the live version. Nav links inside the page point to `/` (dashboard) so `system → project` flow works.

### DIFF
--- a/crates/harness-server/src/handlers/mod.rs
+++ b/crates/harness-server/src/handlers/mod.rs
@@ -6,6 +6,7 @@ pub mod gc;
 pub mod health;
 pub mod learn;
 pub mod observe;
+pub mod overview;
 pub mod preflight;
 pub mod projects;
 pub mod rules;
--- a/crates/harness-server/src/handlers/overview.rs
+++ b/crates/harness-server/src/handlers/overview.rs
@@ -11,7 +11,7 @@ use crate::http::AppState;
 use axum::{extract::State, http::StatusCode, Json};
 use chrono::{DateTime, Duration, Timelike, Utc};
 use harness_core::types::{Decision, Event, EventFilters};
-use harness_observe::{quality::QualityGrader, stats};
+use harness_observe::quality::QualityGrader;
 use serde_json::{json, Value};
 use std::collections::HashMap;
 use std::sync::Arc;
@@ -30,7 +30,14 @@ const FEED_LIMIT: usize = 40;
 /// GET /api/overview — JSON payload driving the system overview page.
 pub async fn overview(State(state): State<Arc<AppState>>) -> (StatusCode, Json<Value>) {
     let now = Utc::now();
-    let since_dt = now - Duration::hours(OVERVIEW_WINDOW_HOURS);
+    // Snap the query window to the start of the oldest bucket on the hour
+    // axis so that SQL rows and JS buckets agree. Without this, tasks
+    // completed in the partial "now - 24h .. oldest-bucket-start" slice are
+    // counted in the global `merged_24h` KPI but dropped when their
+    // `strftime('%H:00:00Z')` hour key falls outside the axis, which makes
+    // per-project sums understate the KPI for every request that is not
+    // exactly on the hour.
+    let since_dt = window_start(now);
     let since = since_dt.to_rfc3339();
 
     // ---- global task queue counts (reuse existing services) ----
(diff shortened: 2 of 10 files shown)


majiayu000:
@codex

majiayu000:
Addressed all three Codex findings in e5ff459 (DCO-signed):

- **P1** /api/overview fetch now sends `Authorization: Bearer` from `sessionStorage.harness_token` (same flow as dashboard.js), and a 401 triggers the same token overlay. Pages fed from the deployments with `api_token` configured will now show live data.
- **P2** `compute_rule_fail_rate_pct` uses all `rule_check` events as denominator with blocks as numerator, so the metric is no longer pinned to 0% / 100%.
- **P3** SQL `since` and `build_hour_axis` both anchor to `hour_top(now) - (THROUGHPUT_BUCKETS-1)h` via a shared `window_start` helper, so global `merged_24h` equals the sum of per-project throughput buckets for any request time.

Thanks for the review. @codex please take another look.

majiayu000:
Fixed in d6123a6 (DCO-signed): wrapped both sides of the `updated_at` predicate in `datetime(...)` so SQLite normalises the stored `YYYY-MM-DD HH:MM:SS` against the RFC3339 `since` before comparing. Same-day rows are no longer dropped.

Added `count_done_since_counts_same_day_rows` as a regression — it inserts a done task, queries with `since = now - 1h` and asserts count=1 (would return 0 without the fix). Locks down both `count_done_since` and `done_per_project_hour_since`.

Thanks @codex — good catch.

majiayu000:
Round 2 addressed in 9ccb5fb (DCO-signed):

- **P2-A** Restored `idx_tasks_status_updated` range scan. Changed `count_done_since` and `done_per_project_hour_since` signatures from `&str` to `DateTime<Utc>`, and format `since` with a new `sqlite_datetime()` helper that matches SQLite's `datetime('now')` output exactly (`YYYY-MM-DD HH:MM:SS`). Raw `updated_at >= ?` compare is now monotonic so the planner can range-scan, and the same-day regression test still passes.
- **P2-B** Orphan project buckets: after draining registered projects, remaining `per_project_buckets` entries (deleted projects with recent done tasks, or tasks never attributed to a project) are emitted as fallback series — `unregistered · <key>` or `unassigned` — so the stacked-area sum always matches `merged_24h`.

Thanks @codex for the index catch — that was hiding a real perf regression behind correct numbers.

majiayu000:
Fixed in b919e14 (DCO-signed). Good catch — valid point that neither `dashboard.js` nor `dashboard.rs` parses `?project=`, so the query param was a no-op and the subtitle was promising scope that didn't exist.

Until the dashboard grows actual project scoping (tracked separately), rows now navigate to plain `/` and the subtitle reads "click a row to open the dashboard — per-project scope is tracked as a follow-up". Matches server behaviour rather than making a false promise.

@codex please take another look.

PR #857, opened 2026-04-19
fix(web): align useTasks hook to /api/tasks for consistency

### LINKED ISSUE #844: fix(web): useTasks hook likely missing /api prefix on tasks endpoint (Gemini feedback from #836)
Gemini's review on merged PR #836 noted:

> Review feedback highlights a likely missing /api prefix in the new tasks endpoint

The feedback was written in the review body, no inline comment, so harness's review loop did not flag it as a blocker and I merged on CI-green. A follow-up sweep is needed.

## Action

- Open the new \`useTasks\` hook added in PR #836 (likely under \`web/src/\` or \`crates/harness-server/static/\`)
- Verify the fetch URL against the mounted axum routes in \`crates/harness-server/src/http.rs\` (list_tasks is at \`/tasks\` inside the router, but the router itself may be mounted under \`/api\`)
- If the prefix is missing, the dashboard fetches silently fail (404) and the Kanban board stays empty

## Verify

- Open dashboard in browser; DevTools Network tab should show 200 on \`/api/tasks\` (or whatever the correct path is)
- If 404: fix the useTasks URL

majiayu000:
## Summary

- Verified that the server's `GET /tasks` route exists and is functional (no 404)
- Added `GET /api/tasks` alias in `http_router.rs` so the web frontend can use the consistent `/api/*` namespace (matching `/api/dashboard`, `/api/overview`)
- Updated `useTasks` hook in `queries.ts` to fetch `/api/tasks` instead of `/tasks`

The `/tasks` route is preserved for backward compatibility with agents that POST/GET tasks directly.

Closes #844

## Test plan

- [x] `cargo fmt --all` — clean
- [x] `cargo test --workspace --lib --exclude harness-server --exclude harness-observe --exclude harness-workflow` — all pass
- [x] `bun run vitest run` in `web/` — 44/44 pass


majiayu000:
Closing — this branch was created before today's DB migration PRs (#829, #831, #851) merged. The branch's files list includes `crates/harness-core/src/db.rs`, `db_pg.rs`, and the store modules, which on merge would revert today's Postgres work back toward the SQLite state. Safer to re-issue against current `main` once the remaining review queue clears. Original intent preserved in the linked issue.

PR #858, opened 2026-04-19
fix(db): PgMigrator SQL splitter + JSONB columns + optimistic locking enforcement

### LINKED ISSUE #845: fix(db): PgMigrator SQL splitter + missing JSONB + incomplete optimistic locking (Gemini feedback from #829)
Gemini's review on merged PR #829 flagged three real correctness concerns that were not addressed before merge:

## 1. Fragile SQL statement splitting in PgMigrator

\`PgMigrator::apply\` (crates/harness-core/src/db_pg.rs) splits each migration's SQL on literal \`;\` characters. Any migration that embeds a semicolon in a string literal, PL/pgSQL function body, or dollar-quoted block will crash mid-migration.

**Fix**: use a SQL-aware splitter or require one statement per Migration entry.

## 2. JSON stored as TEXT, not JSONB

The migration ports existing SQLite TEXT columns as-is. Postgres offers JSONB with native indexing and operators. Columns that store JSON (e.g. task_artifacts content, task_prompts payload, anywhere we read .request_settings) should be JSONB.

**Fix**: change column types to JSONB in new migrations and update read paths.

## 3. Optimistic locking partially implemented

The \`version\` column was added for optimistic locking but:
- Update queries are missing the \`WHERE version = \$expected\` clause
- Read paths don't return the current version to callers

**Fix**: audit every update query on \`tasks\` and add \`WHERE version = ? AND tasks.id = ?\` with \`UPDATE ... SET version = version + 1\`.

## Source

Gemini review body on PR #829 (body-only, no inline comments):
> Key issues identified include a fragile SQL statement splitting method in the migrator, the need for native JSONB types for JSON columns, and incomplete implementation of optimistic locking, specifically missing column mappings and logic checks in the update queries.

majiayu000:
## Summary

- **SQL splitting correctness**: Replace naive `split(';')` in `PgMigrator::apply()` with a proper SQL-aware splitter (`pg_split_statements`) that handles semicolons inside single-quoted strings, double-quoted identifiers, line comments (`--`), block comments (`/* */`), and dollar-quoted PL/pgSQL blocks (`$$...$$`, `$tag$...$tag$`)
- **JSONB column types**: Add migration v19 to convert the `rounds`, `depends_on`, `phase`, and `request_settings` TEXT columns to JSONB; update all SELECT queries to use `col::text AS col` casts so sqlx can decode them as `String`, and update INSERT/UPDATE to bind the raw JSON string (PostgreSQL auto-parses it into JSONB)
- **Optimistic locking enforcement**: Add `version` field to `TaskState` and `TaskRow`; extend the UPDATE query with `AND version = $N` + `version = version + 1`; check `rows_affected() == 0` and return a conflict error — previously the `version` column was incremented but never checked, making the locking a no-op

## Test plan

- [x] `cargo check --workspace --all-targets` — clean
- [x] `RUSTFLAGS="-Dwarnings" cargo check --workspace --all-targets` — clean (no warnings)
- [x] `cargo fmt --all -- --check` — passes
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — passes
- [x] `cargo test -p harness-core` — 317 tests pass (includes 7 new `pg_split_statements` unit tests)
- [x] `cargo test -p harness-agents` — 89 tests pass
- [x] Unit tests for `pg_split_statements`: single statement, two statements, semicolon-in-string, semicolon-in-line-comment, semicolon-in-block-comment, dollar-quoted block, v17 migration shape

Closes #845

### DIFF
--- a/crates/harness-core/src/db_pg.rs
+++ b/crates/harness-core/src/db_pg.rs
@@ -47,6 +47,167 @@ pub async fn pg_open_pool_schematized(database_url: &str, schema: &str) -> anyho
     Ok(pool)
 }
 
+/// Split a SQL string into individual statements, correctly handling:
+/// - Single-quoted string literals (including escaped `''`)
+/// - Double-quoted identifiers (including escaped `""`)
+/// - Line comments (`--`)
+/// - Block comments (`/* */`)
+/// - Dollar-quoted strings (`$$...$$` or `$tag$...$tag$`)
+fn pg_split_statements(sql: &str) -> Vec<String> {
+    let mut statements: Vec<String> = Vec::new();
+    let mut current = String::new();
+    let mut chars = sql.chars().peekable();
+
+    while let Some(ch) = chars.next() {
+        match ch {
+            // Line comment: -- ... \n
+            '-' if chars.peek() == Some(&'-') => {
+                current.push(ch);
+                if let Some(c) = chars.next() {
+                    current.push(c);
+                }
+                for c in chars.by_ref() {
+                    current.push(c);
+                    if c == '\n' {
+                        break;
+                    }
+                }
+            }
+            // Block comment: /* ... */
+            '/' if chars.peek() == Some(&'*') => {
+                current.push(ch);
+                if let Some(c) = chars.next() {
+                    current.push(c);
+                }
+                loop {
+                    match chars.next() {
+                        None => break,
+                        Some(c) => {
+                            current.push(c);
+                            if c == '*' && chars.peek() == Some(&'/') {
+                                if let Some(c) = chars.next() {
+                                    current.push(c);
+                                }
+                                break;
+                            }
+                        }
+                    }
+                }
+            }
+            // Single-quoted string: '...' with '' escape
+            '\'' => {
+                current.push(ch);
+                loop {
+                    match chars.next() {
+                        None => break,
+                        Some(c) => {
+                            current.push(c);
+                            if c == '\'' {
+                                if chars.peek() == Some(&'\'') {
+                                    if let Some(q) = chars.next() {
+                                        current.push(q);
+                                    }
+                                } else {
+                                    break;
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            // Double-quoted identifier: "..." with "" escape
+            '"' => {
+                current.push(ch);
+                loop {
+                    match chars.next() {
+                        None => break,
+                        Some(c) => {
+                            current.push(c);
+                            if c == '"' {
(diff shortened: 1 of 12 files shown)


majiayu000:
🔴 **Blocking: migration v19 fails on the JSONB cast because `rounds` has a non-castable TEXT default**

From the Test job log on run 24636986208:

\`\`\`
---- event_replay::tests::replay_skips_phantom_task_not_in_db stdout ----
Error: migration v19 'convert JSON columns from TEXT to JSONB' failed:
error returned from database: default for column "rounds" cannot be cast automatically to type jsonb
[sql: ALTER TABLE tasks ALTER COLUMN rounds TYPE JSONB USING rounds::jsonb]
\`\`\`

At least 4 tests hit this same migration error (`replay_skips_phantom_task_not_in_db`, `replay_event_log_has_pr_url_checkpoint_has_none`, `replay_and_recover_integration`, `replay_terminal_failed_overrides_implementing`) — the migration step is hard-failing on every test that exercises the v19 path.

## Root cause

The `rounds` column currently has a TEXT default (probably `''` or `'[]'` as literal TEXT). Postgres refuses to cast the column to JSONB while the column still has a TEXT-typed default:

> default for column "rounds" cannot be cast automatically to type jsonb

This is Postgres's safety net against accidentally dropping valid defaults during a type migration.

## Fix

Split the ALTER into three steps so the default is handled explicitly:

\`\`\`sql
-- In migration v19, for each column being converted:
ALTER TABLE tasks ALTER COLUMN rounds DROP DEFAULT;
ALTER TABLE tasks ALTER COLUMN rounds TYPE JSONB USING rounds::jsonb;
ALTER TABLE tasks ALTER COLUMN rounds SET DEFAULT '[]'::jsonb;
\`\`\`

Same three-step pattern must be applied to any other TEXT→JSONB conversion in v19 that also has a column default. Check the migration text for every `ALTER COLUMN ... TYPE JSONB` and add the DROP DEFAULT / SET DEFAULT pair around each.

## Extra

- Run the failing tests locally after the fix with `DATABASE_URL` pointing at a local Postgres: `cargo test -p harness-server event_replay`
- The four tests above should all pass; they're standing in as your regression net for v19's correctness

Once the migration is fixed and CI is green, the other two concerns from #845 (SQL-aware splitter, optimistic locking) still need review. I didn't look at those yet — if you make me, I'll read the full diff once the migration is unblocked.

majiayu000:
Closing — rebase requires manual conflict resolution; please re-open as a fresh PR if still relevant

PR #870, opened 2026-04-20
fix(server): return 501 from password_reset instead of false success

### LINKED ISSUE #868: [P1] U-17: password_reset handler returns false success — no email is ever sent
## Description

`POST /auth/reset-password` (added in the recent `misc_routes.rs` refactor) responds with `{"status":"ok","message":"If that email is registered, a reset link has been sent."}` but the handler contains no email-sending logic. The endpoint is rate-limited and logs a hashed email, but no reset link is dispatched to the user. This is a declaration-execution gap: the response message makes a factual claim the code does not fulfil.

## Affected File

`crates/harness-server/src/http/misc_routes.rs:524-574`

```rust
pub(crate) async fn password_reset(...) -> (StatusCode, Json<serde_json::Value>) {
    // ... rate limiting, logging ...
    (
        StatusCode::OK,
        Json(json!({"status": "ok", "message": "If that email is registered, a reset link has been sent."})),
    )
    // ← no email is sent
}
```

## Risk

Users who request a password reset receive a confident success message but no action is taken. The feature is silently broken. If the endpoint is exposed in any UI (the route is registered in the router and exempt from Bearer-auth), users will be misled into believing their reset request was processed.

## Recommended Fix

Choose one:
1. **Implement** actual email delivery using an SMTP or transactional email service, wiring through a config field (e.g., `server.smtp_*`) and returning an error when unconfigured.
2. **Gate** the endpoint: return `501 Not Implemented` (or remove the route) until email sending is implemented, so callers know the feature is absent.
3. **Document** in a TODO comment + return `202 Accepted` with a message that does not claim an email was sent (e.g., `"password reset not yet implemented"`), so the gap is explicit.

majiayu000:
## Summary

- `POST /auth/reset-password` no longer claims that a reset email was sent when no email delivery exists
- The disabled endpoint still validates input, enforces the configured per-address rate limit, and logs a hashed email for auditability
- Focused password-reset tests now cover the disabled 501 contract, blank-email rejection, normalized-email throttling, and auth exemption without depending on the DB-backed state builder
- `Cargo.lock` now bumps `rustls-webpki` to `0.103.13` to clear `RUSTSEC-2026-0104`

## Root Cause

The handler response drifted away from the real implementation twice: first by reporting a successful reset without sending mail, and later by over-correcting into an unthrottled stub that removed the abuse controls on an auth-exempt route.

## Fix

Keep the endpoint disabled with `501 Not Implemented` until SMTP/transactional email delivery is wired up, but preserve email normalization, per-address throttling, and hashed-email audit logging around that temporary contract.

Closes #868


majiayu000:
@codex

PR #871, opened 2026-04-20
fix(server): return 501 for unimplemented password reset endpoint

majiayu000:
## Summary

- `POST /auth/reset-password` previously responded `200 OK` with the message _"If that email is registered, a reset link has been sent."_ despite containing zero email-sending logic
- Replace the misleading handler with `501 Not Implemented` and an honest error message
- Remove the now-unused `PasswordResetRequest` struct and dead rate-limiter call

## Test plan

- [ ] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [ ] `cargo test --workspace` passes (pre-existing DB pool failures in `harness-observe` are unrelated)
- [ ] Manual: `curl -X POST /auth/reset-password` returns `501` with `{"error":"password reset is not yet available"}`

Signed-off-by: majiayu000 <1835304752@qq.com>

### DIFF
--- a/crates/harness-server/src/http/misc_routes.rs
+++ b/crates/harness-server/src/http/misc_routes.rs
@@ -511,65 +511,13 @@ pub(crate) async fn ingest_signal(
     )
 }
 
-#[derive(serde::Deserialize)]
-pub(crate) struct PasswordResetRequest {
-    pub(crate) email: String,
-}
-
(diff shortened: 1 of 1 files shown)


majiayu000:
Closing as duplicate of #870 (opened 1 minute earlier, fixes same issue #868 with more complete body including `Closes #868`). Tracking in #870.

PR #901, opened 2026-04-22
Use stored PR URLs for startup recovery validation

majiayu000:
This change fixes startup recovery for pending tasks that already have a stored PR URL.

What changed:
- use the stored PR URL directly when validating recovered PR-backed tasks with `gh pr view`
- keep URL parse validation, but avoid converting to `owner/repo#number`
- add a focused unit test for the argument shape

Why:
- repo-qualified refs are brittle for `gh pr view` and can be interpreted as branch refs
- using the persisted PR URL is the most faithful recovery input after restart

Validation:
- `cargo fmt --all`
- `cargo check -p harness-server`
- `cargo test -p harness-server recovered_pr_view_args_use_stored_pr_url -- --test-threads=1`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2985,9 +2985,9 @@ dependencies = [
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.12"
+version = "0.103.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


majiayu000:
@codex

PR #908, opened 2026-04-22
fix(intake): recover when sprint planner emits dependency outside sprint plan

### LINKED ISSUE #884: fix(intake): recover when sprint planner emits dependency outside sprint plan
## Context

Today the repo sprint intake for `majiayu000/loom` aborted with:

```text
intake: invalid sprint DAG — aborting repo="majiayu000/loom" error=issue 19 depends on 16 which is not in the sprint plan
```

This is not a runtime task failure; it is a planner-output validation failure. The current behavior aborts the entire sprint when the planner emits a dependency on an issue that is not present in the same sprint plan.

## Facts

- `validate_dag()` explicitly rejects any dependency edge whose upstream issue is not in `all_task_issues`:
  - `issue {issue} depends on {dep} which is not in the sprint plan`
  - [source: `crates/harness-server/src/intake/mod.rs:191-197`]
- Intake aborts immediately on this validation error:
  - `tracing::error!(repo, error = %e, "intake: invalid sprint DAG — aborting");`
  - [source: `crates/harness-server/src/intake/mod.rs:360-361`]
- The planner prompt currently allows emitting `depends_on` edges across issues, but intake does not repair or downgrade invalid edges before validation.

## Problem

A single bad dependency edge from the sprint planner causes the **entire repo sprint** to abort, even when:
- the referenced upstream issue is already done in a previous sprint, or
- the planner simply forgot to include the upstream in this plan, or
- the downstream issue could still be run standalone after dropping the invalid edge.

This creates a brittle failure mode where planner output quality directly determines whether any work starts.

## Desired behavior

Before validating the DAG, intake should normalize planner output so the system degrades safely instead of aborting the whole sprint.

### Preferred repair order

1. If `depends_on` references an issue not in `all_task_issues`, check whether that upstream issue is already terminal/done in the tracker/store.
   - If yes, drop the edge and log an info/warn event.
2. If the upstream is not in-plan and not already done, either:
   - auto-add the upstream issue to the sprint plan, **or**
   - drop the downstream issue from the plan with an explicit skip reason, **or**
   - strip the invalid edge and mark the downstream task as needing clarification.
3. Only abort the sprint if the planner output is unrecoverably invalid after normalization.

## Minimal acceptable fix

- Do **not** abort the whole sprint on the first out-of-plan dependency.
- Convert this case into a recoverable planner-output normalization step.
- Emit structured logs describing what was repaired or skipped.

## Acceptance criteria

- [ ] Intake no longer aborts the entire sprint solely because one task depends on an out-of-plan issue
- [ ] Unit tests cover:
  - missing upstream already completed
  - missing upstream not completed
  - multiple mixed valid/invalid edges
- [ ] Logs clearly report whether the edge was dropped, the task was skipped, or the upstream was injected
- [ ] `cargo test --workspace` green

## Notes

This issue is about **intake robustness**, not planner quality. Even if the planner should ideally avoid emitting invalid dependencies, intake should not fail-stop the entire batch on one malformed edge.

majiayu000:
## Summary
- normalize sprint-plan dependencies before DAG validation so intake no longer aborts the full repo sprint on one out-of-plan edge
- drop missing edges when the upstream issue is already done in the task store, and skip only the affected task when the upstream is still unresolved
- add intake unit coverage for completed upstream, unresolved upstream, and mixed valid/invalid dependency graphs

Closes #884

## Validation
- cargo fmt --all
- cargo test -p harness-server normalize_sprint_tasks_
- cargo test --workspace *(blocked by unrelated harness_observe event_store tests failing with Postgres session exhaustion / pool timeout: EMAXCONNSESSION)*

### DIFF
--- a/crates/harness-server/src/intake/mod.rs
+++ b/crates/harness-server/src/intake/mod.rs
@@ -389,6 +389,47 @@ fn parse_issue_external_id(external_id: &str) -> Option<u64> {
         .ok()
 }
 
+fn latest_timestamp_at_least(candidate: Option<&str>, existing: Option<&str>) -> bool {
+    match (candidate, existing) {
+        (Some(candidate), Some(existing)) => candidate >= existing,
+        (Some(_), None) => true,
+        (None, Some(_)) => false,
+        (None, None) => true,
+    }
+}
+
+fn latest_unblocking_issue_numbers(
+    external_statuses: std::collections::HashMap<
+        String,
+        (Option<String>, crate::task_runner::TaskStatus),
+    >,
+) -> std::collections::HashSet<u64> {
+    let mut latest_by_issue: std::collections::HashMap<
+        u64,
+        (Option<String>, crate::task_runner::TaskStatus),
+    > = std::collections::HashMap::new();
+
+    for (external_id, (created_at, status)) in external_statuses {
+        let Some(issue) = parse_issue_external_id(&external_id) else {
+            continue;
+        };
+        let should_replace = latest_by_issue
+            .get(&issue)
+            .map(|(existing_created_at, _)| {
+                latest_timestamp_at_least(created_at.as_deref(), existing_created_at.as_deref())
+            })
+            .unwrap_or(true);
+        if should_replace {
+            latest_by_issue.insert(issue, (created_at, status));
+        }
+    }
+
+    latest_by_issue
+        .into_iter()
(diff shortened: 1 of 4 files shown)


majiayu000:
@codex

majiayu000:
@codex

majiayu000:
@codex

PR #910, opened 2026-04-22
Unify task ownership and recovery under a single scheduler authority

### LINKED ISSUE #898: Unify task ownership and recovery under a single scheduler authority
## Problem

Harness has strong persistence primitives, but it does not have one clearly authoritative scheduler state.

Today orchestration truth is spread across multiple places, including:

- task rows / task cache
- task checkpoints
- task event replay log
- runtime host leases / runtime state snapshot
- project cache / registry state
- background recovery/redispatch queues

This makes restart and partial-failure behavior harder to reason about than it should be.

## Why this matters

When orchestration truth is fragmented, the system has to reconcile several partially durable views of reality:

- which tasks exist
- which tasks are claimed
- which tasks are actually running
- which tasks are retryable
- which host/worker owns a task
- which recovery path should apply

That increases the chance of:

- split-brain ownership
- stale claims or leases
- retries that do not correspond to a live owner
- recovery paths that depend on the order of replay/rebuild steps
- operator confusion about what the system believes is happening

## Desired design

Move toward a single authoritative scheduler model.

That does not necessarily mean “one table only”, but it does mean the system should have one canonical ownership/retry/recovery state machine instead of several partially overlapping ones.

Examples of acceptable directions:

1. define one scheduler authority that owns claims, retries, and live/ recovering ownership
2. reduce startup recovery layering by making replay/checkpoint/lease semantics converge into one execution authority
3. unify runtime-host lease state with task execution state instead of treating them as adjacent systems
4. expose one operator-visible notion of task ownership / run generation / recovery generation

## Expected behavior

- it is always clear which subsystem is authoritative for task ownership
- runtime host leases cannot drift independently from task execution state
- restart recovery no longer depends on several loosely ordered channels agreeing after boot
- APIs/dashboard can explain ownership and recovery state from one scheduler truth

## Acceptance criteria

- task ownership, recovery ownership, and retry ownership are modeled through one authoritative scheduler contract
- runtime host lease semantics are explicitly integrated with task lifecycle semantics
- restart flow documents and enforces a single source of truth for “what is running / claimed / recoverable now”
- operator-visible state can identify the current owner/generation of a task without consulting multiple subsystems

## Related

- Related: #891
- Related: #895

majiayu000:
Closes #898

## Summary
- move task ownership, recovery, and retry authority onto persisted task scheduler state
- drive runtime-host claims through the task row and stop persisting lease ownership in runtime snapshots
- expose scheduler authority on task DTOs and derive runtime-host lease pressure from task state

## Validation
- `cargo fmt --all`
- `cargo check`
- `cargo test --workspace` (fails under concurrent agents because the shared Postgres backend hits `(EMAXCONNSESSION) max clients reached in session mode - max clients are limited to pool_size: 15`)


majiayu000:
Rebased onto current main (includes #919, #934, #925).

Post-rebase fixes applied:
- Added missing struct fields (`failure_kind`, `workspace_path`, `workspace_owner`, `run_generation`) to task constructors in tests — these fields were added to the Task struct by #919
- Fixed `scheduler` field in workspace task fixtures from #919
- Fixed import ordering in `task_runner/state.rs` to satisfy `cargo fmt`
- All commits now have DCO `Signed-off-by: majiayu000 <1835304752@qq.com>`

Note: `cargo test --workspace` still fails for DB-dependent tests due to EMAXCONNSESSION pool limit in the shared Postgres backend — this is a pre-existing environment constraint, not introduced by this PR.

majiayu000:
@codex review

PR #920, opened 2026-04-23
Give review and planner tasks a dedicated lifecycle

### LINKED ISSUE #897: Give review-only and planner tasks a dedicated lifecycle instead of the implement path
## Problem

`periodic_review` and `sprint_planner` are semantically review/planner tasks, but they currently execute through the generic implement path.

That means they typically:

- enter `Implementing`
- run one prompt turn
- either become `Done` directly or fail as generic implementing work

They do not have a dedicated lifecycle/status model that matches what the system is actually doing.

## Why this is a problem

This mismatch causes multiple operational problems:

1. UI/dashboard/operator expectations do not match persisted status
2. review-only work can disappear from review-oriented views after restart
3. restart recovery treats them like generic prompt work instead of review/planner work
4. future logic additions keep getting bolted onto the implement pipeline with source-specific conditionals

This is a design smell: task semantics are being forced through the wrong state machine.

## Current behavior

Today review-only/planner sources are detected by source strings such as:

- `periodic_review`
- `sprint_planner`

and then special-cased inside the implement path.

That means there is no first-class status/phase model for:

- scheduler-generated review work
- planner synthesis work
- their restart-safe checkpoints
- their dedicated completion semantics

## Desired design

Give review-only and planner-only tasks their own lifecycle contract instead of treating them as implementation work.

Examples:

- dedicated task kinds
- dedicated statuses, phases, or sub-phases
- dedicated checkpoint payloads
- dedicated startup redispatchers

Possible status direction (illustrative only):

- `review_generating`
- `review_waiting`
- `planner_generating`
- `planner_waiting`
- `recovering_review`
- `recovering_planner`

The exact names are flexible. The key requirement is to stop routing these tasks through the generic implement path.

## Acceptance criteria

- `periodic_review` no longer completes through the generic implement-only special case
- `sprint_planner` no longer completes through the generic implement-only special case
- review-only/planner-only tasks have a dedicated lifecycle contract in code
- restart recovery preserves these tasks as review/planner work instead of flattening them into generic `implementing` semantics
- task summaries/API/dashboard can distinguish these tasks from issue implementation work

## Related

- Related: #891
- Related: #895
- Related: #896

majiayu000:
Closes #897

## Summary
- persist first-class task kinds and restart-safe system input for trusted review/planner prompt tasks
- dispatch review/planner work through dedicated non-implementation lifecycles and recover them into kind-specific waiting states
- expose task kind and new statuses through the task APIs and dashboard/worktree UI

## Verification
- cargo check
- cargo test (with DATABASE_URL=postgresql://postgres:postgres@127.0.0.1:55439/postgres against local Postgres)
- bun run typecheck
- bun run test -- src/routes/dashboard/Active.test.tsx

### DIFF
--- a/crates/harness-server/src/handlers/operator_snapshot.rs
+++ b/crates/harness-server/src/handlers/operator_snapshot.rs
@@ -274,6 +274,7 @@ mod tests {
     fn missing_timestamps_serialize_as_null() {
         let stalled_task = crate::task_runner::TaskState {
             id: harness_core::types::TaskId("stalled-task".to_string()),
+            task_kind: crate::task_runner::TaskKind::Issue,
             status: crate::task_runner::TaskStatus::Implementing,
             turn: 1,
             pr_url: None,
@@ -295,6 +296,7 @@ mod tests {
             triage_output: None,
             plan_output: None,
             request_settings: None,
+            system_input: None,
         };
         let stalled_json = stalled_task_json(&stalled_task);
 
@@ -398,6 +400,7 @@ mod tests {
         for i in 0..25u32 {
             let mut task = crate::task_runner::TaskState {
                 id: harness_core::types::TaskId(format!("fail-task-{i}")),
+                task_kind: crate::task_runner::TaskKind::Issue,
                 status: crate::task_runner::TaskStatus::Failed,
                 turn: 1,
                 pr_url: None,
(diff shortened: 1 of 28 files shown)


majiayu000:
@codex

PR #942, opened 2026-04-25
fix(store): replace dead owner-guard in claim_for_runtime_host with explicit SchedulerOwnerKind match

### LINKED ISSUE #938: [P1] Logic: dead owner-guard in claim_for_runtime_host masks split-brain risk
## Description

`claim_for_runtime_host` in `crates/harness-server/src/task_runner/store.rs` (added in #910) contains a safety check at line 471 that is dead code due to the preceding `clear_to_queued()` call:

```rust
// store.rs:465-473
if entry.scheduler.owner.is_some() && entry.scheduler.has_live_runtime_host_lease(now) {
    return Ok(None);  // correct: already claimed by live runtime host
}
if entry.scheduler.owner.is_some() && !entry.scheduler.has_live_runtime_host_lease(now) {
    entry.scheduler.clear_to_queued();  // clears owner → owner is now None
}
if entry.scheduler.owner.is_some() {   // DEAD CODE: always false after clear_to_queued()
    return Ok(None);
}
```

`clear_to_queued()` unconditionally sets `self.owner = None`. The intent of the third guard — "don't steal a task from a non-RuntimeHost owner" — is silently lost.

## Impact

**Currently not exploitable** because "Pending + Scheduler owner" is not a reachable state today: `claim_scheduler` is only called from inside `update_status`, which atomically transitions status away from `Pending` at the same time. However:

1. The dead code communicates a false invariant to maintainers — the guard appears to protect against stealing Scheduler-owned tasks but does nothing.
2. Any future code path that sets a Scheduler owner on a Pending task (e.g., a recovery flow or a two-phase commit pattern) will bypass the intended safety check silently, enabling a runtime host to steal a task that the local scheduler believes it owns.
3. The two-phase nature of the claim (update DashMap, then persist to DB) means a runtime host that successfully claims the task after the local scheduler's `clear_to_queued()` erasure will write a conflicting row to the database, which the local scheduler's subsequent `persist` will overwrite — leaving the DB and in-memory state diverged.

## File/Line Reference

`crates/harness-server/src/task_runner/store.rs:468-473`

## Recommended Action

Replace the current three-block chain with an explicit check on the owner kind:

```rust
if let Some(owner) = &entry.scheduler.owner {
    match owner.kind {
        SchedulerOwnerKind::RuntimeHost => {
            if entry.scheduler.has_live_runtime_host_lease(now) {
                return Ok(None);  // live lease, back off
            }
            entry.scheduler.clear_to_queued();  // stale lease, reclaim
        }
        SchedulerOwnerKind::Scheduler => {
            return Ok(None);  // local scheduler has it, never steal
        }
    }
}
```

This makes the intended invariant explicit and ensures any future "Pending + Scheduler owner" state is handled correctly.

majiayu000:
## Summary

- The three-block chain in `claim_for_runtime_host` (store.rs:465–473) contained a dead third guard: `clear_to_queued()` sets `owner = None` unconditionally, making the subsequent `owner.is_some()` always false and silently dropping the "don't steal a Scheduler-owned task" invariant.
- Replaced with a single `if let Some(owner) / match owner.kind` block that explicitly handles `RuntimeHost` (live lease → back off, stale lease → reclaim) and `Scheduler` (always return `None`).
- Added two regression tests covering both branches of the new guard.

## Test plan

- [ ] `claim_for_runtime_host_blocked_by_scheduler_owner` — runtime host cannot steal a Scheduler-owned Pending task
- [ ] `claim_for_runtime_host_reclaims_stale_runtime_host_lease` — stale RuntimeHost lease is correctly reclaimed by a new host
- [ ] Existing `claim_for_runtime_host_rolls_back_cache_when_persist_fails` still covers the DB rollback path
- [ ] `cargo clippy --workspace --all-targets -- -D warnings` clean

Closes #938

### DIFF
--- a/crates/harness-server/src/task_runner/store.rs
+++ b/crates/harness-server/src/task_runner/store.rs
@@ -462,14 +462,18 @@ impl TaskStore {
             return Ok(None);
         }
         let original_scheduler = entry.scheduler.clone();
-        if entry.scheduler.owner.is_some() && entry.scheduler.has_live_runtime_host_lease(now) {
-            return Ok(None);
-        }
-        if entry.scheduler.owner.is_some() && !entry.scheduler.has_live_runtime_host_lease(now) {
-            entry.scheduler.clear_to_queued();
-        }
-        if entry.scheduler.owner.is_some() {
-            return Ok(None);
+        if let Some(owner) = &entry.scheduler.owner.clone() {
+            match owner.kind {
+                super::state::SchedulerOwnerKind::RuntimeHost => {
+                    if entry.scheduler.has_live_runtime_host_lease(now) {
+                        return Ok(None);
+                    }
+                    entry.scheduler.clear_to_queued();
+                }
+                super::state::SchedulerOwnerKind::Scheduler => {
+                    return Ok(None);
+                }
+            }
         }
 
         entry.scheduler.claim_runtime_host(host_id, expires_at);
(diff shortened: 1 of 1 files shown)


majiayu000:
Closing as obsolete: superseded by merged #940 (make Scheduler-owner protection explicit in claim_for_runtime_host). The same code area has already been addressed on main.

PR #962, opened 2026-04-27
fix(workflow): replace update_project_path with repair_project_id in sweeper

majiayu000:
## Summary
- Replaces `update_project_path` call in the feedback sweeper (`background.rs`) with `repair_project_id`, which correctly rekeyes the row primary key in a single transaction
- Adds a warning doc comment on `update_project_path` explaining the stale-id hazard
- Fixes U-11: stale primary key after project-path patch caused duplicate workflow rows and lost workflow state history

## Root Cause
`update_project_path` patches only `data::jsonb->>'project_id'` but the row `id` is derived from `workflow_id(project_id, repo, issue)`. After the patch, any downstream call recomputes the id from the new canonical path, misses the stale row, and inserts a new placeholder via `ON CONFLICT(id) DO NOTHING` — silently discarding all prior state.

## Fix
`repair_project_id` (already implemented at `issue_lifecycle.rs:831`) does a correct delete-then-insert in a transaction, keeping the primary key consistent with the new project_id.

## Test plan
- [x] `cargo fmt --all` passes
- [x] `cargo clippy --workspace --all-targets -- -D warnings` passes
- [x] `cargo test --workspace` passes (pre-existing DB pool exhaustion failures in `plan_db` and integration tests are unrelated to this change; the three directly relevant tests `update_project_path_patches_project_id_field`, `update_project_path_noop_on_missing_id`, and `repair_project_id_rekeyes_row` all pass)


majiayu000:
Superseded by #961 — same fix (sweeper uses repair_project_id) merged at 2026-04-27. Closing as duplicate to avoid double work.

PR #1003, opened 2026-04-28
Backfill partial workflow migrations

majiayu000:
## Summary
- Continue workflow migration when the target schema is non-empty instead of skipping the whole copy.
- Preserve existing target rows while backfilling missing legacy issue and project workflow rows.
- Add regression coverage for partial target schemas in both issue and project workflow migrations.

## Verification
- cargo fmt --all
- cargo check
- HARNESS_DATABASE_URL=postgres://postgres:postgres@localhost:55431/harness_test cargo test --package harness-server --lib http::builders::registry::tests:: -- --test-threads=1
- HARNESS_DATABASE_URL=postgres://postgres:postgres@localhost:55431/harness_test cargo test --workspace

Stacked on #1001; retarget to main after #995 and #1001 land.


majiayu000:
Closing as superseded by #1014. The clean #1014 branch is now based on main and includes the partial-backfill behavior without overwriting existing target workflow rows.

PR #1014, opened 2026-04-28
Fix workflow backfill overwrite

majiayu000:
## Summary
- Backfill missing legacy issue/project workflow rows even when the target schema already contains rows.
- Preserve existing target rows by inserting legacy rows only when absent.
- Add regression coverage for partial-target backfills so canonical target state is not overwritten.

## Validation
- cargo fmt --all
- cargo check --message-format short
- cargo test -p harness-server migration_backfills_partial_target -- --test-threads=1

Closes #1003


majiayu000:
@codex

PR #1016, opened 2026-04-28
refactor: add agent adapter execution strategy

majiayu000:
## Summary
- Add an adapter execution strategy to distinguish control-only adapters from turn-executing adapters.
- Register Claude as control-only and Codex as turn-executing, removing name-based turn lifecycle branching.
- Add semantic JSON-RPC error helpers and apply them to rule/check validation paths.
- Document the next design cleanup direction and clarify rule-enforcement comments.

## Validation
- `cargo fmt --all`
- `cargo check`
- `DATABASE_URL=postgres://harness:harness@localhost:5432/harness cargo test -- --test-threads=1`

Note: a parallel `cargo test` run exposed existing state interference in two webhook route tests; the full suite passes serially with the local database URL above.

### DIFF
--- a/crates/harness-agents/src/registry.rs
+++ b/crates/harness-agents/src/registry.rs
@@ -5,13 +5,39 @@ use harness_core::{
 use std::collections::HashMap;
 use std::sync::Arc;
 
+/// Describes how an optional interactive adapter participates in turn execution.
+///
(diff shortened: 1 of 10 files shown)


majiayu000:
Closing as superseded: the adapter execution strategy cleanup is already on main via #1019, and this older branch is stale/dirty with a Clippy failure.

PR #1026, opened 2026-04-29
Return queued task responses under saturation

### LINKED ISSUE #841: bug(server): POST /tasks blocks indefinitely when concurrency queue is saturated
## Symptom

Sending `POST /tasks` with a single-task body (`{"issue": N, ...}`, `{"pr": N, ...}`, or `{"prompt": "...", ...}`) hangs the HTTP handler **indefinitely** when the global task queue is full. `curl -m 60` times out on the client; the server's TCP connection remains in ESTABLISHED state on its side, holding a parked future. Multiple such hung requests accumulate and appear to starve the accept loop (see companion issue on listener death).

Reproduce (against a freshly started server with `server.max_concurrent_tasks = 8`):

```sh
# Seed queue to 8/8 running + 5 queued via any means (real tasks or prompt jobs)
# Then:
time curl -s -m 60 -X POST http://localhost:9800/tasks \
  -H "Content-Type: application/json" \
  -d '{"prompt": "ping", "repo": "majiayu000/harness",
       "project": "/abs/path/to/harness"}'
# → curl exits with 28 (CURLE_OPERATION_TIMEDOUT) at 60.00s, no body.
```

Meanwhile `/health`, `/projects/queue-stats`, `/api/overview` still respond normally — only the POST-tasks handler is stuck.

## Root cause

`crates/harness-server/src/http/task_routes.rs:278` in `enqueue_task`:

```rust
// Acquire concurrency permit before spawning. Blocks if all slots are
// occupied; rejects immediately if the waiting queue is full.
let permit = state
    .concurrency
    .task_queue
    .acquire(&project_id, req.priority)
    .await
    .map_err(|e| EnqueueTaskError::Internal(e.to_string()))?;
```

The comment says "rejects immediately if the waiting queue is full" but the implementation does not. `TaskQueue::acquire()` is an unbounded async wait — the handler parks on the semaphore until a slot frees.

Contrast with the batch handler, which does the right thing. `task_routes.rs:386-397` on `enqueue_task_background`:

> Unlike `enqueue_task`, this function never blocks on concurrency permit acquisition. The task is registered with Pending status right away, and a background tokio task waits for a slot and then begins execution. This keeps the `/tasks/batch` HTTP handler responsive even when all concurrency slots are occupied.

So the async-enqueue path exists, but only `POST /tasks/batch` uses it. `POST /tasks` stayed on the old blocking path.

## Impact

1. Any operator submitting single tasks during a saturation spike sees requests hang. They retry, creating more parked futures.
2. Parked futures hold their request-handler task, which still owns resources (parsed JSON body, DB connections from the pool, etc.).
3. Combined with orphan futures from client timeouts (see companion issue), sustained saturation eventually breaks the listener.

## Fix options

**A. Use the existing async path for single POST too.** Factor the shared pre-work out of `enqueue_task_background` into a helper, then have `enqueue_task` call it and return the task id immediately with `status="pending"`. Most operators already expect `{"status":"running", "task_id":"..."}` to mean "accepted, will run when slots free".

**B. Add `?async=true` or `?nonblocking=true` query param.** Keeps backward compat for clients that rely on the blocking semantics (if any) while giving a way out.

**C. Bounded-wait semantics.** Add a bounded waiting-queue size on `TaskQueue::acquire()`. When the wait queue is full, return 503 with `Retry-After` so clients back off instead of hanging. The skill docs say the comment already advertises this — the fix could just wire it up.

Recommend **A** as the default behaviour — matches `/tasks/batch`, keeps the HTTP handler unconditionally fast, and makes both endpoints semantically identical modulo shape.

## Acceptance criteria

- [ ] `POST /tasks` returns within ~100 ms even when all slots are full (measure with `time curl` under sustained load).
- [ ] Returned `task_id` is observable in `GET /tasks/{id}` with `status="pending"` until a slot frees.
- [ ] An integration test seeds the queue to 8/8, then asserts 10 more `POST /tasks` calls each complete in < 500 ms and produce distinct task ids.
- [ ] Documentation comment on `enqueue_task` (or its replacement) accurately describes the behaviour.

majiayu000:
## Summary
- return accepted queued responses from POST /tasks and /tasks/batch without waiting for task permits
- move permit acquisition into background execution for HTTP task submission
- keep queue admission failures visible by marking preregistered tasks failed
- keep preregistered queued tasks visible to active duplicate detection by preserving project roots
- cover saturated single, batch, repeated, and duplicate-external-id submissions

Closes #841

## Tests
- cargo fmt --all
- cargo fmt --all -- --check
- cargo check
- HARNESS_DATABASE_URL=postgres://harness:harness@localhost:5432/harness HARNESS_DATABASE_POOL_MAX_CONNECTIONS=8 HARNESS_DATABASE_POOL_ACQUIRE_TIMEOUT_SECS=30 cargo test -p harness-server --lib create_task_deduplicates_external_id_while_queued_under_saturation -- --nocapture
- HARNESS_DATABASE_URL=postgres://harness:harness@localhost:5432/harness HARNESS_DATABASE_POOL_MAX_CONNECTIONS=8 HARNESS_DATABASE_POOL_ACQUIRE_TIMEOUT_SECS=30 cargo test -p harness-server --lib task_runner::spawn::tests::register_pending_task_keeps_repo_metadata_visible -- --exact --nocapture
- HARNESS_DATABASE_URL=postgres://harness:harness@localhost:5432/harness HARNESS_DATABASE_POOL_MAX_CONNECTIONS=8 HARNESS_DATABASE_POOL_ACQUIRE_TIMEOUT_SECS=30 cargo test -p harness-server --lib saturat -- --nocapture
- HARNESS_DATABASE_URL=postgres://harness:harness@localhost:5432/harness HARNESS_DATABASE_POOL_MAX_CONNECTIONS=8 HARNESS_DATABASE_POOL_ACQUIRE_TIMEOUT_SECS=30 cargo test -p harness-server --lib http::tests::health_degraded_when_subsystem_missing -- --exact --nocapture
- HARNESS_DATABASE_URL=postgres://harness:harness@localhost:5432/harness HARNESS_DATABASE_POOL_MAX_CONNECTIONS=8 HARNESS_DATABASE_POOL_ACQUIRE_TIMEOUT_SECS=30 cargo test -p harness-server --lib
- RUSTFLAGS=\"-Dwarnings\" cargo clippy --workspace --all-targets
- RUSTFLAGS=\"-Dwarnings\" cargo check --workspace --all-targets

### DIFF
--- a/crates/harness-server/src/http/task_routes.rs
+++ b/crates/harness-server/src/http/task_routes.rs
@@ -58,15 +58,6 @@ pub(crate) async fn enqueue_task_background_in_domain(
         .await
 }
 
-fn create_task_status_response(state: &AppState, task_id: &task_runner::TaskId) -> &'static str {
-    match state.core.tasks.get(task_id) {
-        Some(task) if matches!(task.status, task_runner::TaskStatus::AwaitingDeps) => {
-            "awaiting_deps"
-        }
-        _ => "running",
-    }
-}
-
 /// A single task entry in the detailed batch format.
 #[derive(Debug, Deserialize)]
 pub struct BatchTaskItem {
@@ -298,18 +289,15 @@ pub(super) async fn create_task(
     State(state): State<Arc<AppState>>,
     Json(req): Json<task_runner::CreateTaskRequest>,
 ) -> Response {
-    match enqueue_task(&state, req).await {
-        Ok(task_id) => {
-            let status = create_task_status_response(state.as_ref(), &task_id);
-            (
-                StatusCode::ACCEPTED,
-                Json(json!({
-                    "task_id": task_id.0,
-                    "status": status
-                })),
-            )
-                .into_response()
-        }
+    match enqueue_task_background(state, req, None).await {
+        Ok(task_id) => (
+            StatusCode::ACCEPTED,
+            Json(json!({
+                "task_id": task_id.0,
+                "status": "queued"
+            })),
+        )
+            .into_response(),
         Err(EnqueueTaskError::BadRequest(error)) => {
             (StatusCode::BAD_REQUEST, Json(json!({ "error": error }))).into_response()
         }
--- a/crates/harness-server/src/http/tests.rs
+++ b/crates/harness-server/src/http/tests.rs
@@ -8,9 +8,10 @@ use harness_core::{
 };
 use hmac::{Hmac, Mac};
 use sha2::Sha256;
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
-use tokio::sync::Mutex;
+use tokio::sync::{Mutex, Notify, Semaphore};
 use tower::ServiceExt;
 
 struct CapturingAgent {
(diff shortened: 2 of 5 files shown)


majiayu000:
Addressed the Gemini review feedback on the current branch:

- `register_pending_task` now preserves `req.project` in `TaskState::project_root`, so active duplicate detection can see queued tasks before they acquire a permit
- added `create_task_deduplicates_external_id_while_queued_under_saturation` to prove duplicate submissions return the existing queued task id
- kept the existing repeated prompt test distinct because prompt-only submissions without an external id should remain separate tasks
- expanded `wait_for_task_project_root` to 200 x 50ms to avoid a 500ms CI flake window

PR #1061, opened 2026-05-05
Fix release-plz token fallback

majiayu000:
## Summary
- Use the repository GitHub token when RELEASE_PLZ_TOKEN is not configured.
- Apply the fallback to checkout and release-plz action authentication.

## Verification
- ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release-plz.yml"); puts "release-plz.yml parses"'
- cargo fmt --all
- cargo test --workspace (fails only on harness-core DB tests because HARNESS_DATABASE_URL is not configured)


majiayu000:
@codex review

PR #1064, opened 2026-05-07
Default workflow runtime kind to claude_code

majiayu000:
## Summary

- WORKFLOW.md's `runtime_kind` / `runtime_profile` are independent from `[agents] default_agent` in the server config — repos without a project-level WORKFLOW.md were forced to `codex_jsonrpc` / `codex-default` even when the operator set `default_agent = "claude"`.
- Switch the compiled defaults to `claude_code` / `claude-default`. Repos that still need codex can override via their own WORKFLOW.md.
- Move the `workflow.rs` test module to `workflow_tests.rs` via `#[path]` so the file stays under the U-16 800-line ceiling and future default-value tweaks are not blocked by the size guard.
- Flip harness's own `WORKFLOW.md` from `codex_jsonrpc` to `claude_code` so the harness repository is consistent with the new defaults.

## Why

This unblocks single-source-of-truth agent selection for users who only want to set `default_agent = "claude"` once. Before this change, every repo without a custom WORKFLOW.md silently fell back to codex regardless of the operator's intent.

## Test plan

- [x] `cargo test -p harness-core --lib config::workflow` — 7/7 pass with new default-case asserts
- [x] `RUSTFLAGS="-Dwarnings" cargo check --workspace --all-targets` (fresh target dir) — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] Local smoke: server reload, `repo_backlog` poll activity dispatches to `claude --model opus` against majiayu000/harness, prompt packet shows `"runtime_kind": "claude_code"`, `"runtime_profile": "claude-default"`, and the matching `claude` child process is observed


majiayu000:
Follow-up tracking the activity-result failure observed after switching to `claude_code`: #1065. The routing change in this PR is independent and correct on its own; #1065 is the diagnostic + downstream fix to make the new default actually produce activity output.

PR #1068, opened 2026-05-07
refactor(validation): introduce ValidationExecutor trait

majiayu000:
## Summary

Add a \`ValidationExecutor\` trait in \`harness-core::validation\` so the production sites that spawn \`Command::new(\"sh\")\` directly can be substituted with a mock in tests.

\`PostExecutionValidator\` is wired through the trait:

- New field \`validation_executor: Arc<dyn ValidationExecutor>\`, defaulting to \`ShellValidationExecutor\` in the public constructors so existing call paths compile unchanged.
- New \`with_executor()\` builder for tests that want to inject a mock and assert how the validator interprets specific outcomes (timeout, non-zero exit, captured stderr).
- \`run_command(&self, ...)\` now delegates to the executor; error message wording is preserved so log greps continue to match.

## Why

Two production sites historically spawned \`sh -c\` directly, bypassing every trait abstraction. That made it impossible to unit-test how the merge gate or the post-execution validator reacts to specific shell outcomes without spawning real processes. The \`ValidationExecutor\` trait gives both sites a mockable substitution point. This PR wires the post-validator through it; the second site (\`run_test_gate\`) follows in a sibling PR (see *Out of scope* below).

## Behaviour preservation

- \`ShellValidationExecutor\` preserves the existing \`kill_on_drop(true)\` plus piped-stdout/stderr semantics.
- All 26 \`post_validator::tests\` pass unchanged, including the timeout-kills-child invariant.
- 5 new unit tests in \`harness_core::validation::tests\` exercise the shell executor for success, non-zero exit, captured stdout/stderr, timeout, and environment propagation.

## Out of scope (deferred)

- The second \`sh -c\` site at \`task_executor::run_test_gate\`. Wiring the trait through it requires threading an \`executor\` parameter through \`run_review_loop\` (already 30+ parameters) and \`task_executor::run_task\`, large enough to deserve its own PR.
- Holding the executor on \`AppState\`. The trait is mockable today via \`PostExecutionValidator::with_executor\`; AppState wiring becomes necessary only once the test-gate site also takes the trait, so it lands together with that follow-up.

## Test plan

- [x] \`cargo test -p harness-core --lib validation::\` — 5 tests pass
- [x] \`cargo test -p harness-server --lib post_validator\` — 26 tests pass
- [x] \`RUSTFLAGS=\"-Dwarnings\" cargo check --workspace --all-targets\` passes
- [x] \`cargo fmt --all -- --check\` passes
- [ ] Gemini code review bot
- [ ] CI Result green

### DIFF
--- a/crates/harness-core/src/lib.rs
+++ b/crates/harness-core/src/lib.rs
@@ -14,6 +14,7 @@ pub mod shell_safety;
 mod test_support;
 pub mod tool_isolation;
 pub mod types;
+pub mod validation;
 
 pub use config::misc::OtelExporter;
 pub use types::{
--- a/crates/harness-core/src/validation.rs
+++ b/crates/harness-core/src/validation.rs
@@ -12,8 +12,20 @@ use async_trait::async_trait;
 use std::path::Path;
 use std::process::Stdio;
 use std::time::Duration;
+use tokio::io::{AsyncRead, AsyncReadExt};
 use tokio::process::Command;
 use tokio::time::timeout;
+
+/// Per-stream cap on captured validation output. Verbose validators (e.g.
+/// `cargo test -- --nocapture`, `npm test`) can produce hundreds of MiB of
+/// output that the caller never reads in full — only error excerpts and exit
+/// codes. Bounding each stream prevents an unbounded `wait_with_output()`
+/// buffer from inflating harness memory or OOM-killing the server.
+///
+/// 256 KiB is enough to retain the head of typical build errors and is small
+/// enough that even pathological 1 GiB log streams cost at most ~512 KiB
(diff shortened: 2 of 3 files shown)


majiayu000:
Reviewed. I would not merge this yet. The validation abstraction is useful and the normal CI jobs are passing, but the current `ShellValidationExecutor` uses `wait_with_output()`, so arbitrary configured validators can buffer unbounded stdout/stderr in memory. For verbose validators such as `cargo test -- --nocapture` or `npm test`, that can inflate harness memory or kill the process even when the validator succeeds.

Please stream validator output or cap captured output before merging, then rerun CI.

PR #1069, opened 2026-05-07
refactor(task-kind): centralise per-kind metadata on TaskKind

majiayu000:
## Summary

Consolidate the genuinely-repeated \`match task_kind\` blocks in \`harness-server\` into either inherent methods on \`TaskKind\` or file-local helpers:

- \`TaskKind::orphan_recovery_failure_reason()\` replaces the per-kind failure-reason mapping in \`http::background::orphan_recovery_failure_reason\` (which is now a one-line delegate kept for symmetry with the existing call sites).
- \`TaskKind::prompt_task_label()\` replaces the per-kind label match in \`task_runner::request\`. Returns \`Option<&'static str>\` so a future variant can opt out of prompt-only intake instead of being silently coerced to \`\"prompt task\"\`.
- New file-local helper \`runtime_external_id\` in \`http::task_query_routes\` deduplicates two byte-identical \`match task_kind\` blocks.

## Why — and a scope correction

The architecture review flagged this as the highest-leverage refactor (~10 production sites). A closer audit found the situation milder than reported:

- Most \`match task_kind\` sites are unique per-call decisions, not duplicated logic — extracting each into a method would just add indirection.
- \`recovery_queue_domain\` was already a separate free function; nothing to centralise there.
- The few sites that are actually duplicated (\`external_id\`) or that benefit from being expressed as inherent methods (\`orphan_recovery_failure_reason\`, \`prompt_task_label\`) are addressed by this PR.

The result is a smaller refactor than originally planned, but the cases it covers are the ones that actually pay back when adding a new \`TaskKind\` variant.

## Out of scope (intentional)

- The \`match task_kind\` block at \`workflow_runtime_worker.rs:153\` lives on a file with concurrent WIP; deferred to PR-5b once that work lands.
- One-line \`matches!(task_kind, ...)\` patterns in \`task_runner::spawn\` and \`task_executor::helpers\`. Wrapping each in a method costs more readability than it saves.
- The \`description\` mapping in \`task_query_routes::runtime_workflow_task_summary\` mixes workflow-specific data with TaskKind branching; not pure enough to belong on the enum.

## Behaviour preservation

- Per-kind strings are byte-equal copies of the previous inline literals.
- 70 \`task_runner\` tests pass unchanged.

## Test plan

- [x] \`cargo test -p harness-server --lib task_runner\` — 70 tests pass
- [x] \`RUSTFLAGS=\"-Dwarnings\" cargo check --workspace --all-targets\` passes
- [x] \`cargo fmt --all -- --check\` passes
- [ ] Gemini code review bot
- [ ] CI Result green

> Note: an unrelated test \`http::background::tests::runtime_worker_loop_policy_keeps_polling_when_server_root_disables_worker\` is currently failing in this working tree, but the failure also reproduces on \`origin/main\` with this PR\\'s changes stashed. Not caused by this PR.

### DIFF
--- a/crates/harness-server/src/http/background.rs
+++ b/crates/harness-server/src/http/background.rs
@@ -138,13 +138,7 @@ fn task_allows_prompt_orphan_recovery(task: &task_runner::TaskState) -> bool {
 }
 
 fn orphan_recovery_failure_reason(task: &task_runner::TaskState) -> &'static str {
-    match task.task_kind {
-        task_runner::TaskKind::Issue => "orphaned issue task: issue number not persisted",
-        task_runner::TaskKind::Pr => "orphaned PR task: PR number not persisted",
-        task_runner::TaskKind::Prompt => "orphaned prompt-only task: prompt not persisted",
-        task_runner::TaskKind::Review => "orphaned review task: prompt not persisted",
-        task_runner::TaskKind::Planner => "orphaned planner task: prompt not persisted",
-    }
+    task.task_kind.orphan_recovery_failure_reason()
 }
 
 pub(super) fn recovery_queue_domain(task_kind: task_runner::TaskKind) -> task_routes::QueueDomain {
--- a/crates/harness-server/src/http/task_query_routes.rs
+++ b/crates/harness-server/src/http/task_query_routes.rs
@@ -178,11 +178,7 @@ fn runtime_workflow_task_summary(
         .data
         .get("issue_number")
         .and_then(|value| value.as_u64());
-    let external_id = match task_kind {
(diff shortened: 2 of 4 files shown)


majiayu000:
LGTM from the diff. This is a tight refactor: the duplicated `external_id` mapping is moved to a file-local helper, and the two repeated per-kind string mappings become `TaskKind` methods without changing the emitted strings. I do not see a behavior regression in the touched paths.

At review time, `Format`, `Check`, PR description, and harness-pr are passing; Clippy/Test/Security/disabled-module checks are still running, so I would wait for those to finish before merging.

PR #1077, opened 2026-05-07
fix(db): SQL-aware migration splitter + optimistic locking on tasks

### LINKED ISSUE #845: fix(db): PgMigrator SQL splitter + missing JSONB + incomplete optimistic locking (Gemini feedback from #829)

majiayu000:
## Summary

Closes Gemini's two outstanding correctness concerns from #845:

1. **SQL-aware migration splitter.** `PgMigrator::apply` split each migration's SQL on a literal `;`, so any future migration that embedded a semicolon in a string literal, PL/pgSQL function body, or dollar-quoted block would crash mid-migration. Replaced with `pg_split_statements`, a small state machine that recognizes single-quoted strings (with `''` escape), double-quoted identifiers, `--` line comments, `/* */` block comments, and `$tag$ ... $tag$` dollar-quoted blocks. 13 unit tests cover each construct.
2. **Optimistic locking actually enforced.** The `tasks.version` column existed (migration v18) but no UPDATE used it as a precondition (`grep -rn 'WHERE version' crates/` returned 0 hits before this PR). `TaskDb::update` now reads the expected version off the loaded `TaskState`, sends `WHERE id = $N AND version = $M`, increments inside the same UPDATE, and surfaces a typed conflict error when `rows_affected == 0`. The in-memory `TaskStore` cache mirrors the increment after every successful persist / runtime-host claim / runtime-host release so the next write supplies a matching expected version.

Out of scope per the [issue's narrowing comment](https://github.com/majiayu000/harness/issues/845#issuecomment-3019380000): JSONB column migration. `exec_plans.data` is already JSONB (#861); the remaining TEXT-encoded JSON columns on `tasks` are intentionally left as TEXT here.

### Refactors driven by the 800-line per-file ceiling

- `crates/harness-core/src/db_pg.rs`: tests and the new splitter live in sibling files (`db_pg_tests.rs`, `db_pg_split.rs`).
- `crates/harness-server/src/task_db/queries_tasks.rs`: startup-recovery write paths (`apply_replayed_state`, `recover_in_progress`) move to `queries_recovery.rs`.

## Test plan

- [x] `cargo fmt --all -- --check` clean
- [x] `cargo check --workspace --all-targets` clean
- [x] `RUSTFLAGS=-Dwarnings cargo check --workspace --all-targets` clean
- [x] `cargo test -p harness-core --lib db_pg_split` — 13/13 pass
- [x] `cargo test -p harness-server --lib` — 1158/1160 pass; the two failures are the pre-existing Supabase-pool contention flake under parallel test load (`skill_governance_history_limit_respected`, `invalid_workflow_schema_namespace_disables_optional_workflow_stores`), both pass in isolation
- [x] `cargo test -p harness-core --tests` clean
- [x] `cargo test -p harness-workflow --tests` clean

Closes #845


majiayu000:
@codex review

PR #1079, opened 2026-05-07
feat(http/shutdown): staged drain-then-force shutdown with hard deadline

### LINKED ISSUE #843: feat(http/shutdown): staged drain-then-force shutdown with visible progress (replaces unbounded graceful_shutdown wait)
## Problem

The current shutdown path in `crates/harness-server/src/http/mod.rs:187-193` uses `axum::serve(...).with_graceful_shutdown(shutdown_signal())` unconditionally, which means:

- Ctrl+C triggers `shutdown_signal()` to return
- axum then waits **indefinitely** for every in-flight HTTP connection to close
- Any hung SSE stream, stale dashboard WebSocket, or long-poll handler holds the process forever
- Second Ctrl+C, SIGTERM, and friendly signals are ignored because the signal handler already consumed the first one

Observed today (2026-04-19): harness was running with 8 active tasks across several projects. Ctrl+C appeared to do nothing. `kill <pid>` (SIGTERM) also appeared to hang because it hit the same `shutdown_signal` branch that awaits `graceful_shutdown`. The only reliable escape was `kill -9`, which loses log flush, runtime_state snapshot, and any unwritten observability events.

## Goals

1. **Fast feedback**: Ctrl+C should be visibly acknowledged within a second — no silent hang
2. **No lost work**: task checkpoints must still be flushed before the process exits
3. **Bounded wait**: a hard upper bound so the server cannot hang forever
4. **Escalation path**: second Ctrl+C → force; third → unrecoverable exit
5. **Background ≠ request**: long-running agent tasks should never hold HTTP shutdown hostage

### Four-phase shutdown state machine

| Phase | Trigger | Action | User sees |
|-------|---------|--------|-----------|
| **P1 — Stop accepting** | 1st Ctrl+C or SIGTERM | Stop listener accept; mark running tasks `interrupted` in DB; broadcast ws_shutdown | `shutdown: draining N tasks; Ctrl+C again to force (deadline=30s)` |
| **P2 — Drain (bounded)** | P1 complete → up to `drain_timeout_secs` | Wait for in-flight HTTP requests to complete naturally; every 5s log `still draining: M connections, K tasks` | progress updates |
| **P3 — Force** | `drain_timeout_secs` elapsed OR 2nd Ctrl+C | Abort all in-flight handlers; flush event store + runtime_state; close DB pools | `shutdown: force quit, 3 tasks will resume on next start` |
| **P4 — Nuclear** | 3rd Ctrl+C OR hard timeout (`drain_timeout_secs + 30s`) | `std::process::exit(1)` | (no log, immediate exit) |

### Implementation sketch

**`http/mod.rs`**:

```rust
async fn shutdown_signal(drain_deadline: Duration) -> ShutdownReason {
    // P1: wait for first signal
    wait_first_signal().await;
    tracing::info!(
        drain_deadline_secs = drain_deadline.as_secs(),
        \"shutdown: draining; press Ctrl+C again to force\"
    );

    // P2 → P3 transition
    tokio::select! {
        _ = tokio::time::sleep(drain_deadline) => ShutdownReason::DrainTimeout,
        _ = tokio::signal::ctrl_c() => ShutdownReason::UserForced,
    }
}

// In run():
let drain_deadline = config.shutdown.drain_timeout_secs;
let serve = axum::serve(listener, app)
    .with_graceful_shutdown(async move {
        let reason = shutdown_signal(Duration::from_secs(drain_deadline)).await;
        tracing::info!(?reason, \"shutdown: graceful phase ending\");
        ws_shutdown_tx.send(()).ok();
    });

// P4: hard timeout outside graceful_shutdown
let hard_deadline = Duration::from_secs(drain_deadline + 30);
match tokio::time::timeout(hard_deadline, serve).await {
    Ok(Ok(())) => tracing::info!(\"shutdown: clean\"),
    Ok(Err(e)) => tracing::error!(%e, \"shutdown: serve error\"),
    Err(_) => {
        tracing::error!(\"shutdown: hard timeout — process::exit(1)\");
        std::process::exit(1);
    }
}
```

**`task_executor/mod.rs`** (separate but related):

Each turn should check a global `shutdown_flag: Arc<AtomicBool>` before starting agent work. On `true`, persist current phase/turn and return an `ImplementOutcome::Interrupted` variant so the task is re-queueable on next startup.

**`config/default.toml`**:

```toml
[shutdown]
drain_timeout_secs = 30   # time to wait for in-flight HTTP to finish
# Hard exit happens at drain_timeout_secs + 30s regardless
```

### Why this is better than just adding a timeout

- Visibility: the user *sees* what's blocking drain — 5s progress logs tell them which subsystem is hung
- Escalation: no forced \"30s or nothing\" — user can override with 2nd Ctrl+C when they know it's safe
- Durability: interrupted tasks resume instead of being force-killed mid-turn
- Independence: background agents are decoupled from HTTP request lifetime, so they drain on their own timeline without holding axum hostage

## Acceptance

- Pressing Ctrl+C prints a drain message within 100ms
- If no in-flight work: shutdown completes in < 2s
- If in-flight work: shutdown completes within `drain_timeout_secs + 30s` no matter what
- After shutdown, restarting the server resumes any interrupted tasks (status transition `interrupted` → `running`)
- Unit test for `shutdown_signal` with mocked signal stream covers: timeout path, 2nd-ctrl-c path, P4 hard timeout

## Related

- Prior attempt: PR #795 (closed) touched \"prevent spurious shutdown when signal handler install fails\" — partial fix, didn't address the unbounded drain
- #817 introduced completion gates that make task interruption durable — foundation for this work
- Observed 2026-04-19: Ctrl+C appeared unresponsive for > 10 minutes while 8 tasks (vibeguard, auto-run-agent, remem, etc.) were in review loops holding dashboard SSE streams

majiayu000:
## Summary

Replaces the unbounded `axum::serve(...).with_graceful_shutdown(shutdown_signal())` wait in `crates/harness-server/src/http/mod.rs` with a staged state machine driven by a new `ShutdownConfig`. Closes #843.

- **P1** — first SIGINT/SIGTERM logs `shutdown: draining; press Ctrl+C again to force` immediately so Ctrl+C is no longer silent.
- **P2** — drain phase races a configurable `drain_timeout_secs` deadline against a second user signal; whichever fires first ends the graceful phase. Progress is logged every `progress_log_secs` (default 5s) so the user sees what's blocking drain.
- **P3** — the entire `axum::serve` future is wrapped in `tokio::time::timeout(drain_timeout_secs + force_grace_secs, ...)`. If that hard bound is exceeded, the event store is flushed and the process calls `std::process::exit(1)` instead of hanging.

Defaults are conservative: 30s drain, 30s force grace, 5s progress logs.

## Configuration

```toml
[server.shutdown]
drain_timeout_secs = 30   # how long to wait for in-flight HTTP to finish
force_grace_secs = 30     # extra wait before hard process::exit(1)
progress_log_secs = 5     # interval for "still draining" progress logs
```

`ShutdownConfig` lives in `harness-core::config::shutdown` and hangs off `ServerConfig::shutdown`.

## Refactor

Extracted the inline `mod tests` from `crates/harness-core/src/config.rs` into `config/config_test.rs` so the file drops back under the project's 800-line ceiling and the new `pub mod shutdown;` declaration could be added without splitting more code.

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `RUSTFLAGS=\"-Dwarnings\" cargo check --workspace --all-targets`
- [x] `cargo test -p harness-core --lib` — 385 tests pass (includes 3 new `config::shutdown::tests`)
- [x] `cargo test -p harness-server --lib http::shutdown_test` — 4 new tests pass: drain-timeout path, second-signal force, progress-log clamp at 0s, immediately-ready force
- [x] DCO `Signed-off-by` trailer present

## Out of scope

The issue mentions task-executor `ImplementOutcome::Interrupted` plumbing as \"separate but related\". This PR is intentionally limited to the HTTP shutdown surface so the fast-feedback / hard-deadline guarantees ship independently of the task-executor work.

### DIFF
--- a/crates/harness-core/src/config.rs
+++ b/crates/harness-core/src/config.rs
@@ -5,6 +5,7 @@ pub mod misc;
 pub mod project;
 pub mod resolve;
 pub mod server;
+pub mod shutdown;
 pub mod workflow;
 
 use self::agents::*;
--- a/crates/harness-core/src/config/server.rs
+++ b/crates/harness-core/src/config/server.rs
@@ -4,6 +4,7 @@ use std::net::SocketAddr;
 use std::path::PathBuf;
 
 use super::dirs::dirs_data_dir;
+use super::shutdown::ShutdownConfig;
 
 #[derive(Clone, Serialize, Deserialize)]
 pub struct ServerConfig {
@@ -81,6 +82,9 @@ pub struct ServerConfig {
     /// when not configured.
     #[serde(default)]
     pub github_token: Option<String>,
+    /// Staged HTTP shutdown configuration (drain, force, hard-exit deadlines).
+    #[serde(default)]
+    pub shutdown: ShutdownConfig,
 }
 
 impl ServerConfig {
@@ -207,6 +211,7 @@ impl fmt::Debug for ServerConfig {
             password_reset_rate_limit_per_hour,
             constitution_enabled,
             github_token,
+            shutdown,
         } = self;
         f.debug_struct("ServerConfig")
             .field("transport", transport)
@@ -243,6 +248,7 @@ impl fmt::Debug for ServerConfig {
             )
             .field("constitution_enabled", constitution_enabled)
             .field("github_token", &github_token.as_ref().map(|_| "[REDACTED]"))
+            .field("shutdown", shutdown)
             .finish()
     }
 }
@@ -269,6 +275,7 @@ impl Default for ServerConfig {
             password_reset_rate_limit_per_hour: default_password_reset_rate_limit_per_hour(),
             constitution_enabled: true,
             github_token: None,
+            shutdown: ShutdownConfig::default(),
         }
     }
 }
--- a/crates/harness-core/src/config/shutdown.rs
+++ b/crates/harness-core/src/config/shutdown.rs
@@ -1,0 +1,85 @@
+use serde::{Deserialize, Serialize};
+
+/// HTTP server shutdown configuration.
+///
+/// Controls the staged drain-then-force shutdown state machine. The HTTP
+/// listener stops accepting new connections on the first signal, then waits
+/// up to `drain_timeout_secs` for in-flight requests to finish. A second
+/// signal (or the timeout) forces graceful shutdown to end. A hard outer
+/// deadline of `drain_timeout_secs + force_grace_secs` guarantees the
+/// process eventually exits even if the runtime itself is wedged.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ShutdownConfig {
+    /// Seconds to wait for in-flight HTTP requests to drain naturally
+    /// before forcing the graceful shutdown to end. Default: 30.
+    #[serde(default = "default_shutdown_drain_secs")]
+    pub drain_timeout_secs: u64,
+    /// Additional seconds beyond `drain_timeout_secs` before the process
+    /// hard-exits. The hard deadline is `drain_timeout_secs + force_grace_secs`.
+    /// Default: 30.
+    #[serde(default = "default_shutdown_force_grace_secs")]
+    pub force_grace_secs: u64,
+    /// Seconds between drain progress logs while waiting for in-flight
+    /// requests to complete. Default: 5.
+    #[serde(default = "default_shutdown_progress_log_secs")]
+    pub progress_log_secs: u64,
+}
+
+fn default_shutdown_drain_secs() -> u64 {
+    30
+}
+
+fn default_shutdown_force_grace_secs() -> u64 {
+    30
+}
+
+fn default_shutdown_progress_log_secs() -> u64 {
+    5
+}
+
+impl Default for ShutdownConfig {
+    fn default() -> Self {
+        Self {
+            drain_timeout_secs: default_shutdown_drain_secs(),
+            force_grace_secs: default_shutdown_force_grace_secs(),
+            progress_log_secs: default_shutdown_progress_log_secs(),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn defaults_match_documented_values() {
+        let cfg = ShutdownConfig::default();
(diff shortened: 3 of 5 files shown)


majiayu000 (commented):
Blocking review findings:

- [P0] `crates/harness-server/src/http/mod.rs:242-254` starts the hard shutdown timeout when the server starts, not when shutdown starts. `tokio::time::timeout(hard_deadline, serve_future)` wraps the whole `axum::serve(...)` future immediately after bind, so with the default config a normally running server will hit the 30s + 30s timeout and call `process::exit(1)` after about 60 seconds even if no Ctrl+C/SIGTERM was ever received. The hard deadline needs to begin only after the first shutdown signal / force phase starts.

- [P1] `crates/harness-server/src/http/mod.rs:248-299` delays Axum's graceful shutdown signal until after `drain_or_force` completes. `with_graceful_shutdown` only stops accepting new connections once its future resolves, but `staged_shutdown_signal` waits for the drain timeout or a second signal before resolving. That means the advertised drain window is not actually draining; the listener can continue accepting requests during that period, and real graceful shutdown starts only at the end. The first signal should notify Axum immediately, with the drain/force deadline enforced around the actual graceful shutdown.

Additional blocker: this branch is now stale against current `main` and conflicts in `crates/harness-core/src/config.rs`. Current `main` already extracted config tests via `crates/harness-core/src/config_tests.rs` (#1091), while this PR adds `crates/harness-core/src/config/config_test.rs`, so the config-test extraction needs to be rebased/resolved before merge.

Verification performed:
- `git merge --no-commit --no-ff origin/pr/1079` from current `origin/main` -> conflict in `crates/harness-core/src/config.rs`
- `cargo test -p harness-server --lib http::shutdown_test` -> 4 passed
- `cargo test -p harness-core --lib config::shutdown::tests` -> 3 passed

The targeted tests pass, but they only exercise `drain_or_force` and the config parser; they do not cover the serve-level timing where both shutdown regressions occur.

Issue #1127, opened 2026-05-20
Define runtime-native submission identity


majiayu000:
## Problem
`task_id` currently serves two meanings: a legacy task-row identifier and a runtime submission correlation handle. Runtime-owned submissions now also expose `workflow_id`, but API consumers still treat `task_id` as the stable handle for detail, stream, prompt, artifact, proof, cancel, and dashboard flows.

This makes it risky to remove legacy task compatibility because identity semantics are not yet explicit.

## Goal
Define the long-term identity model for runtime-owned submissions. The design should clearly distinguish:

- `workflow_id`: workflow runtime instance identity
- `submission_id` or equivalent: stable external submission handle
- legacy `task_id`: old task-runner row identity, only where a real legacy task exists

## Scope
- Document the API contract for create, list, detail, stream, cancel, proof, artifacts, and prompts.
- Decide whether the public external handle should be named `submission_id` or whether `task_id` remains as a compatibility alias.
- Define migration rules for existing runtime workflow rows with `data.task_id` / `data.task_ids`.
- Add tests that lock the identity contract before implementation PRs start.

## Acceptance Criteria
- A checked-in design doc defines the identity model and migration phases.
- The API compatibility surface is explicit, including any temporary aliases.
- Follow-up implementation issues can be executed without guessing which identifier owns each route.

## Context
Related current behavior:

- `GET /tasks/{id}` falls back from the legacy task store to runtime workflow lookup by task handle.
- Runtime workflow storage indexes `data.task_id` and `data.task_ids`.
- The dashboard still stores `task_id` from `POST /tasks` responses, even when `workflow_id` is present.

This issue should be completed before deleting any legacy task-id compatibility paths.

majiayu000:
Verified this is already implemented by merged PR #1156 (https://github.com/majiayu000/harness/pull/1156), whose merge commit is present on current origin/main. The checked-in contract doc covers workflow_id, submission_id, legacy task_id compatibility, task-compatible API routes, migration phases, and follow-up implementation boundaries, and the contract test still locks the documented surface and runtime alias behavior.\n\nFresh validation from this workflow activity:\n- cargo fmt --all -- --check\n- cargo test -p harness-server --test runtime_submission_identity_contract\n- cargo check -p harness-server --all-targets

PR #1156, opened 2026-05-24
docs(runtime): define submission identity contract

majiayu000:
## Summary

- Add a checked-in runtime submission identity contract for `workflow_id`, `submission_id`, and legacy `task_id` semantics.
- Document the task-compatible API surface for create, list, detail, stream, cancel, proof, artifacts, and prompts.
- Add an integration test that locks the documentation surface and verifies current plus historical runtime submission aliases resolve through `WorkflowRuntimeStore`.

## Validation

- `cargo check -p harness-server --all-targets`
- `cargo test -p harness-server --test runtime_submission_identity_contract`
- `git diff --cached --check`
- `cargo test --workspace --no-fail-fast` was run; it fails locally because this checkout has no `HARNESS_DATABASE_URL`, causing existing DB-backed harness-core/harness-server tests to fail.

Refs #1127


majiayu000:
@codex review

Issue #1114, opened 2026-05-13
refactor: simplify AppState ownership boundaries after service split


majiayu000:
## Summary

The current server state does not show serious `Arc` misuse, but the post-service-split `AppState` still has a few redundant ownership layers that make dependency boundaries harder to read.

This should be treated as a low-priority refactor, not a correctness bug.

## Current evidence

A targeted Clippy pass is clean:

```bash
cargo clippy --workspace --all-targets -- -W clippy::arc_with_non_send_sync -W clippy::mutex_atomic -W clippy::mutex_integer
```

So this is not the `Arc<non-Send/Sync>` class of issue.

The justified `Arc` usage should remain:

- Agent registry entries return shared `Arc<dyn CodeAgent>` / `Arc<dyn AgentAdapter>` handles.
- Task execution moves stores, agents, event stores, queues, and callbacks into spawned Tokio tasks.
- `TaskQueue` uses `Arc<Mutex<_>>` and `Arc<AtomicBool>` for cancellation-safe waiter/permit coordination.

The cleanup candidates are narrower:

- `crates/harness-server/src/http/state.rs`
  - `CoreServices::plan_cache: Arc<DashMap<...>>`
  - `CoreServices::maintenance_active: Arc<AtomicBool>`
  - `NotificationServices::{notification_lagged_total, initializing, initialized}: Arc<Atomic*>`
  - `AppState::{runtime_hosts, runtime_project_cache}: Arc<_>`
- `crates/harness-server/src/http/builders/services.rs`
  - `ServicesBundle::{runtime_hosts, runtime_project_cache}: Arc<_>` mirrors the same boundary.
- `crates/harness-server/src/task_runner/spawn.rs`
  - `let interceptors = Arc::new(interceptors);`
- `crates/harness-server/src/task_executor/{mod.rs,implement_pipeline.rs,review_loop.rs}`
  - task execution passes `Arc<Vec<Arc<dyn TurnInterceptor>>>`; this could be normalized to an `Arc<[Arc<dyn TurnInterceptor>]>` or a small shared interceptor stack type at the service boundary.

## Why this matters

`AppState` itself is already shared as `Arc<AppState>` by the HTTP/router/background-task layer. Extra `Arc` fields inside it are sometimes useful when a subsystem is independently moved into long-lived spawned tasks, but several current fields appear to be read or mutated through `state` rather than independently owned.

[...]

PR #1158, opened 2026-05-24
refactor(execution): split task driver

majiayu000:
## Summary

- Split the oversized task executor driver out of `task_executor/mod.rs` into focused `run_task` and `validation_gate` modules.
- Trimmed the parent module back to orchestration exports and test-only imports.
- Changed interceptor-only helper boundaries in the non-implementation and conflict-gate paths to accept slices instead of `&Arc<Vec<_>>`.

Refs #1114

## Validation

- `git diff --check`
- `cargo check`
- `cargo test -p harness-server task_executor --lib` failed locally because `HARNESS_DATABASE_URL` is not configured; 185 tests passed before the database-backed tests failed.
- `cargo test --workspace --no-fail-fast` failed locally because `HARNESS_DATABASE_URL` is not configured and localhost Postgres is not reachable.


majiayu000:
@codex review

Issue #1129, opened 2026-05-20
Deprecate and remove legacy task-row compatibility paths


majiayu000:
## Problem
Harness still carries compatibility code that maps runtime workflow submissions through legacy `/tasks` semantics. That compatibility cannot be safely removed until runtime-native identity and dashboard/API migration are complete.

## Goal
Remove legacy task-row compatibility paths after runtime-owned flows no longer depend on them.

## Scope
- Inventory every remaining fallback from runtime-owned flows to legacy task rows.
- Remove compatibility routes, response aliases, store fallbacks, and tests that are no longer needed.
- Keep true legacy task-runner functionality only if it is still intentionally supported.
- Add migration or cleanup logic for old persisted rows if required.

## Acceptance Criteria
- Runtime-owned submissions do not require `data.task_id` to masquerade as a legacy task id.
- `/tasks` compatibility behavior is either removed or documented as legacy-only.
- Dead compatibility branches are deleted with regression tests proving runtime-owned flows still work.
- The cleanup plan handles existing persisted runtime workflow rows safely.

## Dependencies
This should start only after the identity contract and runtime-first API/dashboard migration are complete.

PR #1159, opened 2026-05-24
refactor(runtime): persist submission handles

majiayu000:
## Summary
- Split `workflow_runtime_submission.rs` into focused cancel, dependency, and prompt-memory modules so the runtime submission code stays under the project file-size guard.
- Persist an explicit `submission_id` on new runtime issue and prompt submissions.
- Prefer the stable `submission_id` as the public runtime handle while preserving historical `task_id` aliases for existing links and retries.

## Validation
- `cargo fmt --all`
- `git diff --check`
- `git diff --cached --check`
- `cargo check -p harness-server`
- `cargo check`
- `cargo test -p harness-server workflow_runtime_submission --lib`
- `cargo test` still fails locally at the existing database boundary: `harness-core` reports 409 passed and 12 DB tests failed because `HARNESS_DATABASE_URL` is unset. Local `scripts/dev-db.sh` cannot start the DB because Docker Compose v2 is not installed, and `docker`, `psql`, `initdb`, `pg_ctl`, and localhost:5432 are unavailable.

Refs #1129

### DIFF
--- a/crates/harness-server/src/workflow_runtime_submission.rs
+++ b/crates/harness-server/src/workflow_runtime_submission.rs
@@ -1,84 +1,43 @@
-use crate::task_runner::{TaskId, TaskStatus, TaskStore};
+use crate::task_runner::TaskId;
 use harness_workflow::runtime::{
-    activity_result_value_has_closed_issue_evidence, build_issue_submission_decision,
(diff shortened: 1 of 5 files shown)


majiayu000:
@codex review

Issue #1011, opened 2026-04-28
Reduce harness-server test runtime caused by full AppState setup and shared test locks


majiayu000:
## Problem

The Rust test suite is spending several minutes in `harness-server` tests even on a warm build. The slowdown appears to come from full `AppState`/DB setup and shared test locks, not from compilation or from pure business-logic tests.

Measured locally on 2026-04-29:

| Scope | Tests | Time |
| --- | ---: | ---: |
| `cargo test --workspace` | all Rust | `319.12s` |
| `harness_server` lib test binary | 924 | `248.75s` |
| `http::` | 142 | `268.90s` |
| `http::tests::` default parallel | 53 | `148.39s` |
| `http::tests::` single-threaded | 53 | `71.01s` |
| `http::startup_tests::` | 7 | `20.44s` |
| `handlers::` | 99 | `47.05s` |
| `periodic_reviewer::` | 44 | `15.24s` |
| `services::` | 21 | `9.28s` |
| `task_executor::` | 115 | `0.29s` |
| `task_runner::` | 55 | `2.11s` |
| `webhook::` | 30 | `0.01s` |
| `http::task_routes::` | 30 | `0.16s` |

The clearest interference example:

| Scope | Time |
| --- | ---: |
| `http::tests::dispatch_complex_prompt_selects_claude_agent` alone | `0.66s` |
| `http::tests::dispatch_simple_prompt_selects_default_agent` alone | `0.68s` |
| Both dispatch tests with default parallel runner | `106.66s` |
| Both dispatch tests with `--test-threads=1` | `5.04s` |

This suggests default parallel execution is making some tests slower because they contend on shared global/process/DB state.

## Evidence

Relevant code paths:

- `crates/harness-server/src/test_helpers.rs` defines `HOME_LOCK`, which serializes tests that read or mutate process-global `HOME`.
- `crates/harness-server/src/http/init.rs` makes `build_app_state()` acquire a test DB state guard under `#[cfg(test)]`.
- `crates/harness-server/src/test_helpers.rs` stores a DB guard inside test `AppState`, so the lock can live for the whole test.
- Several HTTP and handler tests hold `HOME_LOCK` while building full app state and waiting for async task execution.
- Dashboard handler tests build full state for relatively small response-shape assertions.

The suite also emitted libtest slow-test messages for these paths during a broad run:

- `http::startup_tests::build_app_state_preserves_partial_startup_project_metadata`
[...]

PR #1161, opened 2026-05-24
test(server): narrow AppState DB setup lock lifetime

majiayu000:
## Summary

- Release the test DB setup lock after AppState stores finish opening instead of carrying it for the whole AppState lifetime.
- Apply the narrower lock lifetime to shared server test helpers, HTTP startup construction, stdio tests, and websocket tests.
- Leave oversized http/tests.rs and router/tests.rs untouched because repo guard requires those files to be split before editing.

Refs #1011

## Test plan

- cargo fmt --all
- cargo fmt --all -- --check
- git diff --check
- cargo check
- RUSTFLAGS="-Dwarnings" cargo clippy --workspace --all-targets
- cargo test -p harness-server --lib websocket::tests::is_local_origin_accepts_localhost_variants -- --exact
- cargo test -p harness-server --lib http::background::tests::runtime_worker_loop_policy_clamps_zero_concurrency -- --exact
- cargo test -p harness-server --lib -- --list

## Local test limitation

- cargo test --workspace was attempted but this machine has no HARNESS_DATABASE_URL, so DB-backed tests fail with: database URL is not configured; set server.database_url in TOML or HARNESS_DATABASE_URL in the environment.
- cargo test --workspace -- --skip db::tests:: was also attempted and still failed on DB-backed harness-server tests for the same missing local DB configuration.
- cargo test -p harness-server --lib websocket::tests::websocket_initialize_roundtrip -- --exact --nocapture was attempted and failed for the same missing HARNESS_DATABASE_URL.

### DIFF
--- a/crates/harness-server/src/http/init.rs
+++ b/crates/harness-server/src/http/init.rs
@@ -140,7 +140,7 @@ pub async fn build_app_state(server: Arc<HarnessServer>) -> anyhow::Result<AppSt
     let dir = expand_tilde(&server.config.server.data_dir);
     let project_root = resolve_project_root(&server.config.server.project_root)?;
     #[cfg(test)]
-    let db_state_guard = Some(crate::test_helpers::acquire_db_state_guard().await);
+    let db_setup_guard = crate::test_helpers::acquire_db_state_guard().await;
 
     tracing::debug!(
         data_dir = %dir.display(),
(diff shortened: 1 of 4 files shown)


majiayu000:
@codex review

majiayu000:
Reviewed the Gemini comments. I am treating both as non-actionable:

- `test_helpers.rs` is compiled only through `#[cfg(test)] mod test_helpers` in `crates/harness-server/src/lib.rs`, so there is no production lock acquisition path.
- `_db_state_guard` is only the shared DB setup mutex guard. It does not restore global process state; `HomeGuard` remains the RAII guard for `HOME` restoration, and this change does not alter it.

CI is green and Codex review found no major issues.

Issue #1162, opened 2026-05-24
Track unresolved PR review comments (2026-05-18 to 2026-05-24)


majiayu000:
This tracks PR review threads still marked unresolved for PRs created during 2026-05-18 00:00:00 +08:00 through 2026-05-24 23:59:59 +08:00.

Scope: public, non-archived repositories owned by `majiayu000`.
Source query superset: `user:majiayu000 is:pr created:2026-05-17..2026-05-24 is:public archived:false` filtered to the local week above.
Counting method: GitHub GraphQL `PullRequest.reviewThreads` where `isResolved = false`.

Summary: 11 PRs, 35 unresolved review threads. State split: OPEN 0, MERGED 10, CLOSED 1.

PRs to check:
- [ ] #1130 (MERGED) fix(workflow): commit runtime completion atomically — 2 unresolved thread(s), 1 outdated: https://github.com/majiayu000/harness/pull/1130
- [ ] #1143 (MERGED) fix(runtime): cancel dispatching commands — 1 unresolved thread(s): https://github.com/majiayu000/harness/pull/1143
- [ ] #1146 (CLOSED) Fix atomic PR feedback runtime transitions — 4 unresolved thread(s): https://github.com/majiayu000/harness/pull/1146
- [ ] #1147 (MERGED) fix(runtime): preserve PR-only feedback outcomes — 5 unresolved thread(s): https://github.com/majiayu000/harness/pull/1147
- [ ] #1149 (MERGED) fix(execution): scope synthesized PR feedback task IDs by project — 2 unresolved thread(s): https://github.com/majiayu000/harness/pull/1149
- [ ] #1151 (MERGED) test: use skill helper in store tests — 5 unresolved thread(s), 5 outdated: https://github.com/majiayu000/harness/pull/1151
- [ ] #1153 (MERGED) feat(observe): wire token usage overview KPIs — 3 unresolved thread(s), 1 outdated: https://github.com/majiayu000/harness/pull/1153
- [ ] #1154 (MERGED) feat(web): render paginated task history — 3 unresolved thread(s): https://github.com/majiayu000/harness/pull/1154
- [ ] #1157 (MERGED) [codex] Move task APIs to runtime submission handles — 3 unresolved thread(s), 2 outdated: https://github.com/majiayu000/harness/pull/1157
- [ ] #1160 (MERGED) feat(review): add provider report contract — 5 unresolved thread(s), 1 outdated: https://github.com/majiayu000/harness/pull/1160
[...]

PR #1163, opened 2026-05-24
fix(runtime): address unresolved P1 review followups

majiayu000:
## Summary
- prefer issue-bound workflow instances when resolving PR feedback by PR number
- keep runtime completion and cancellation paths on a command-first lock order
- mark stale runtime jobs for already-terminal workflows according to the terminal state instead of always cancelling them
- guard web test setup localStorage access for opaque-origin environments
- move workflow runtime migrations into a focused module so store.rs stays below the U-16 limit

Refs #1162.

## Verification
- cargo fmt --all -- --check
- CARGO_TARGET_DIR=target/cargo-check cargo check --workspace --all-targets -j 6
- CARGO_TARGET_DIR=target/cargo-check RUSTFLAGS="-Dwarnings" cargo check --workspace --all-targets -j 6
- CARGO_TARGET_DIR=target/cargo-test cargo test -p harness-workflow p1_followups -- --nocapture
- CARGO_TARGET_DIR=target/cargo-test cargo test -p harness-workflow --all-targets -j 6
- npx tsc --noEmit (web)
- npx vitest run (web)

## Note
- cargo test --workspace --all-targets -j 6 was attempted, but existing harness-core db tests failed because HARNESS_DATABASE_URL/server.database_url is not configured in this environment.

### DIFF
--- a/crates/harness-workflow/src/runtime/mod.rs
+++ b/crates/harness-workflow/src/runtime/mod.rs
@@ -14,6 +14,7 @@ pub mod quality_gate;
 pub mod reducer;
 pub mod repo_backlog;
 pub mod store;
+mod store_migrations;
 pub mod submission;
 pub mod validator;
 pub mod worker;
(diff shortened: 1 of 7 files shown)


majiayu000:
@codex review

PR #1326, opened 2026-06-12
Stabilize parallel harness-server tests

### LINKED ISSUE #1191: test(server): remove process-global env and cwd coupling from harness-server tests
Parent: #1011
Related: #1189

## Problem

`harness-server` tests currently require `--test-threads=1` for stable local validation because default parallel execution can contaminate sibling tests through process-global state. That turns the full `harness-server` lib run into a ~399.59s single-threaded wait on the current machine.

Observed shared-state surfaces:

- `crates/harness-server/src/test_helpers.rs` defines `HOME_LOCK` because tests mutate or read process-global `HOME`.
- `crates/harness-server/src/test_helpers.rs` also mutates `HARNESS_DATABASE_URL` through `ensure_test_database_url_override`.
- `crates/harness-server/src/workspace_tests.rs` includes scoped env mutation and current-directory mutation.
- Several handler/router/http tests hold `HOME_LOCK` while building state or exercising async behavior.
- Some test `AppState` values store `_db_state_guard`, extending serialization beyond the narrow setup section.

## Proposed direction

- Pass home/data/project roots explicitly through test helpers instead of mutating `HOME` where possible.
- Pass database URLs explicitly into store constructors instead of mutating `HARNESS_DATABASE_URL` in broad helpers.
- Remove current-directory dependence from workspace tests, or isolate it behind the narrowest possible serial guard.
- Narrow lock lifetimes so tests do not hold global guards while awaiting unrelated async work.
- Leave explicitly serial tests documented, but make the default safe path parallel again.

## Acceptance criteria

- Default parallel `cargo test -p harness-server --lib` is stable for the main server test groups, or remaining serial groups are isolated and documented.
- Tests that only need paths or database URLs do not mutate process-global environment variables.
- `HOME_LOCK` and DB-state locks are held only for the smallest necessary scope.
- The stable local validation path no longer requires serializing all 1321 `harness-server` lib tests.

majiayu000:
## Summary
- Remove the harness-server test helper that mutates HARNESS_DATABASE_URL and route remaining fixtures through explicit cached database URLs.
- Install conservative test-only Postgres pool defaults so parallel test stores do not exhaust local Postgres connections.
- Release the shared HTTP test-state DB setup guard before returning AppState.

Closes #1191

## Validation
- cargo fmt --all -- --check
- cargo check -p harness-server --all-targets
- cargo test -p harness-server --lib skill_governance_history
- cargo test -p harness-server --lib runtime_review_concurrency
- cargo test -p harness-server --lib post_tool_use_returns_violations_when_guard_fires
- cargo test -p harness-server --lib empty_data_dir_produces_empty_project_registry
- /usr/bin/time -p cargo test -p harness-server --lib
- git diff --check

### DIFF
--- a/crates/harness-server/src/http/startup_tests.rs
+++ b/crates/harness-server/src/http/startup_tests.rs
@@ -64,7 +64,6 @@ fn http_listener_starts_before_background_work() {
 async fn persisted_skills_survive_restart() -> anyhow::Result<()> {
     // Hold the shared HOME_LOCK so no sibling test races on HOME.
     let _lock = HOME_LOCK.lock().await;
-    let _ = crate::test_helpers::test_database_url()?;
 
     let sandbox = tempfile::tempdir()?;
     let project_root = sandbox.path().join("project");
--- a/crates/harness-server/src/http/startup_tests.rs
+++ b/crates/harness-server/src/http/startup_tests.rs
@@ -70,7 +70,7 @@ async fn persisted_skills_survive_restart() -> anyhow::Result<()> {
     let project_root = sandbox.path().join("project");
     std::fs::create_dir_all(&project_root)?;
     let data_dir = sandbox.path().join("data");
-    let database_url = crate::test_helpers::ensure_test_database_url_override()?;
+    let database_url = crate::test_helpers::test_database_url()?;
 
     // Redirect HOME to an empty sandbox directory so that
     // $HOME/.harness/skills/ cannot shadow the persisted skill under
--- a/crates/harness-server/src/http/test_fixtures.rs
+++ b/crates/harness-server/src/http/test_fixtures.rs
@@ -141,7 +141,7 @@ async fn make_read_only_route_test_state_with_project_root(
     agent_registry: AgentRegistry,
 ) -> anyhow::Result<Arc<AppState>> {
     let db_state_guard = crate::test_helpers::acquire_db_state_guard().await;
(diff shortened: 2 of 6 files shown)


majiayu000:
Blocking local validation note:

I am not treating `database_pool_max_connections: Some(2)` as merge-ready yet. In the PR worktree, the original `Some(1)` version passed the default parallel `cargo test -p harness-server --lib` run: 1445 passed, 0 failed, finished in 173.00s. After applying the pool=2 suggestion, `cargo test -p harness-server --lib` failed twice with local Postgres pool-acquire timeouts (`1433 passed; 12 failed`), including registry/intake/skills DB-heavy tests. Increasing `HARNESS_DATABASE_POOL_ACQUIRE_TIMEOUT_SECS=120` did not fix it.

Please keep the duplicate `test_database_url()` cleanup, but either revert the test pool default to `Some(1)` or implement a broader connection-budget fix that makes default parallel `cargo test -p harness-server --lib` pass again before merge.

majiayu000:
Addressed the blocking validation note in 7fbbb741 by reverting the harness-server test PostgreSQL pool default back to `Some(1)` while keeping the duplicate `test_database_url()` cleanup.

Validation from this run:
- `cargo fmt --all -- --check` passed
- `cargo test -p harness-server --lib` passed: 1445 passed, 0 failed, finished in 170.56s
- `cargo clippy --workspace --all-targets -- -D warnings` passed
- Remote PR checks are green after the push

PR #1327, opened 2026-06-12
fix(runtime): compact runtime tree polling

### LINKED ISSUE #1165: fix(runtime): make runtime tree safe for dashboard polling
## Problem

The dashboard and `harness status` currently use `GET /api/workflows/runtime/tree` as a real-time state signal. On a live 2026-05-29 Harness run, `/api/workflows/runtime/tree?limit=500&job_limit=1` returned about 315 MB. Two client attempts timed out after 20 seconds with partial 212-221 MB bodies.

The payload is large because the endpoint mixes two use cases:

- a real-time dashboard tree, which only needs counts, status, latest command/job state, and rejected decision hints;
- a full debug dump, which serializes every workflow event, every decision, every command payload, and full runtime job input/output.

Measured against the live server:

- `limit=500&job_limit=0`: 200,804,295 bytes
- `limit=500&job_limit=1`: 315,010,663 bytes
- top node `majiayu000/harness::backlog`: 74 MB by itself, with 7,652 events, 7,559 decisions, and 4,276 commands
- in that node, commands contributed about 34.5 MB, events about 32.0 MB, and decisions about 7.2 MB

This makes the endpoint unsafe for 30-second polling and can mislead operators because the panel may fail to refresh while the runtime is still active.

## Desired design

Make the runtime tree default to a compact polling contract and keep full debug data opt-in:

- add a compact/default response mode that caps rendered commands per workflow and jobs per command;
- include counts such as `event_count`, `decision_count`, `command_count`, and aggregate summary totals without serializing full history;
- return only rejected-decision hints needed by the UI instead of all decision records;
- avoid fetching full runtime event payloads just to compute latest event type or prompt digest;
- add a `summary_only=true` mode for `harness status` so status checks do not download the tree;
- preserve a full debug mode for explicit investigations.

## Acceptance criteria

- `harness status` uses a summary-only runtime query.
- Dashboard runtime polling requests compact data by default.
- The existing full tree details remain available through an explicit query option.
- Tests cover the compact/default shape, summary-only shape, and full-detail compatibility.

majiayu000:
## Summary
- Make `/api/workflows/runtime/tree` default to a compact polling shape with real event/decision/command/job counts, bounded command and job slices, and rejected-decision hints only.
- Preserve the previous full debug payload behind `detail=full` and add `summary_only=true` for status checks.
- Update `harness status` and dashboard polling to use the compact/summary runtime tree contracts.

Closes #1165

## Validation
- `cargo check -p harness-server --all-targets`
- `cargo test -p harness-server runtime_tree -- --nocapture`
- `cargo test -p harness-cli status::tests -- --nocapture`
- `bun run typecheck`
- `bun run test src/routes/dashboard/Active.test.tsx`

### DIFF
--- a/crates/harness-cli/src/commands/status.rs
+++ b/crates/harness-cli/src/commands/status.rs
@@ -139,6 +139,7 @@ async fn get_json(
 fn runtime_tree_path(project_id: Option<&str>, limit: i64) -> String {
     let mut query = form_urlencoded::Serializer::new(String::new());
     query.append_pair("limit", &limit.max(1).to_string());
+    query.append_pair("summary_only", "true");
     if let Some(project_id) = project_id.filter(|value| !value.is_empty()) {
         query.append_pair("project_id", project_id);
     }
@@ -340,7 +341,7 @@ mod tests {
     fn runtime_tree_path_encodes_path_like_project_id() {
         assert_eq!(
             runtime_tree_path(Some("/Users/apple/repo name"), 20),
-            "/api/workflows/runtime/tree?limit=20&project_id=%2FUsers%2Fapple%2Frepo+name"
+            "/api/workflows/runtime/tree?limit=20&summary_only=true&project_id=%2FUsers%2Fapple%2Frepo+name"
         );
     }
 
@@ -348,7 +349,7 @@ mod tests {
     fn runtime_tree_path_clamps_limit_and_encodes_project_id() {
         assert_eq!(
             runtime_tree_path(Some("/project-a"), 0),
-            "/api/workflows/runtime/tree?limit=1&project_id=%2Fproject-a"
+            "/api/workflows/runtime/tree?limit=1&summary_only=true&project_id=%2Fproject-a"
         );
     }
 
--- a/crates/harness-server/src/http/misc_routes.rs
+++ b/crates/harness-server/src/http/misc_routes.rs
@@ -364,7 +364,7 @@ impl WorkflowRuntimeTreePagination {
         let offset = offset.max(0) as usize;
         let total = total.max(0) as usize;
         let next_offset = offset.saturating_add(returned);
-        let has_more = next_offset < total;
+        let has_more = !summary_only && next_offset < total;
         Self {
             limit,
             offset,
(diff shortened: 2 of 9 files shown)


majiayu000:
Additional review finding before merge:

`summary_only=true` currently returns `workflows: []` and `returned: 0`, but the pagination metadata is still built with the real workflow total. For a non-empty project this can produce `has_more: true` with `next_offset: 0`, which can make a generic JSON paginator loop on the same URL.

Please update the summary-only shape so pagination is terminal (`has_more: false`, `next_offset: null`) or otherwise make `WorkflowRuntimeTreePagination` treat `summary_only` as non-pageable. Add a regression assertion for `has_more` and `next_offset` in the summary-only runtime tree test.

majiayu000:
Addressed the current review feedback in `7d6139fc`:

- `summary_only=true` pagination is now terminal (`has_more=false`, `next_offset=null`) while preserving the aggregate totals.
- Compact command rendering now preserves non-object command payloads instead of converting them to `{}`.
- Added runtime tree regression coverage for both cases.

Local validation from this run:
- `cargo check -p harness-server --all-targets`
- `cargo test -p harness-server runtime_tree -- --nocapture`
- `git diff --check`

majiayu000:
Addressed the new Codex review thread in `743eb3f2`:

- Runtime workflow node badges now use `rejected_decision_count` when the compact API provides it, falling back to the rendered rejected-decision list only for older/full payloads.
- The dashboard test now covers a compact payload with one rendered rejected decision but six total rejected decisions.
- The workflow runtime command payload type now permits non-object JSON because the server preserves primitive command payloads.

Additional local validation:
- `bun run typecheck`
- `bun run test src/routes/dashboard/Active.test.tsx`
- `git diff --check`

majiayu000:
Addressed the latest Codex eval-artifact compatibility thread in `930fcfd3`:

- `scripts/evaluate_pr_repair.sh` now requests `/api/workflows/runtime/tree` with `detail=full` when collecting runtime-tree artifacts, preserving runtime job `input`, `output`, and `activity_result_envelope` fields for artifact merging.

Additional local validation:
- `bash -n scripts/evaluate_pr_repair.sh`
- `git diff --check`

PR #1332, opened 2026-06-12
Fix workspace pool leasing for concurrent tasks

### LINKED ISSUE #947: fix(workspace): per-project worktree pool + per-task lease to eliminate same-repo collisions
## Problem

Same-repo parallel tasks share one worktree directory, so two tasks running at the same time on `majiayu000/harness` collide and at least one fails.

Today this surfaces as 12+ failures with `WorktreeCollision: workspace path "..." already in use`. The `harness-delegate` skill (`references/gaps.md` PREF-1) explicitly forces same-repo tasks to run **serially** in user-facing batch flows because of this.

## Desired behavior

Replace the implicit "one worktree per project" with an explicit pool:

- Each project owns a small worktree pool sized by config (default `max_concurrent`).
- A task acquires a worktree from the pool with `lease(task_id)` and releases on completion / failure / cancel.
- A pool entry tracks `last_used_at` and is reset (clean checkout, not deleted) between tasks for fast reuse.
- If the pool is fully leased, new tasks wait on a per-project semaphore instead of crashing the dispatcher.
- The lease metadata is persisted (e.g. `workspace_leases` table) so a server restart can detect orphans and reset them.

## Out of scope

- Removing per-task isolation. Worktrees are still per-task while leased; pool just amortises creation/teardown.
- Cross-project pooling. Each project keeps its own pool.

## Acceptance criteria

- [ ] Two same-repo tasks running concurrently each get a distinct worktree from the pool.
- [ ] No `WorktreeCollision` errors in fresh runs across the active task batch.
- [ ] Pool size is configurable per project under `[[projects]]`.
- [ ] On restart, orphaned leases (process gone) are released automatically.
- [ ] Tests: concurrent acquire/release, restart with orphan lease, pool exhaustion waits instead of failing.
- [ ] Skill memory `harness-delegate PREF-1` can be relaxed once this lands.

## Related

- `crates/harness-server/src/workspace.rs` — current single-path WorkspaceManager.
- `crates/harness-server/src/task_runner/spawn.rs` — where worktree is created today.
- `crates/harness-server/src/task_db/migrations.rs` — likely needs a new `workspace_leases` migration.

majiayu000:
## Summary
- add a WorkspaceManager-backed per-project workspace pool with slot leases
- persist workspace_leases state and release leases on terminal, cancel, retry cleanup, and startup orphan reconciliation
- reset reusable worktree slots before assigning them to a new task while keeping active workspace telemetry intact

Closes #947

## Validation
- cargo fmt --all -- --check
- cargo check -p harness-server --all-targets
- cargo test -p harness-server workspace -- --nocapture
- cargo test -p harness-server runtime_worker_workspace -- --nocapture
- cargo test -p harness-server parallel_dispatch -- --nocapture
- cargo test (failed in tests/notification_delivery.rs: four tests timed out opening the Postgres runtime pool for project_registry while HARNESS_DATABASE_URL is unset)


majiayu000:
Blocking review finding before merge:

P1: cancellation still appears to release the workspace lease after the Tokio task is aborted, but without explicit proof that the real agent process/process group has been killed and waited. The normal cleanup path in `crates/harness-server/src/task_runner/spawn.rs` is skipped on abort, and the watcher still releases the workspace after observing the cancelled join. The existing regression test uses an in-memory blocking agent, so it proves lease timing for the mock future, not that old CLI descendants can no longer mutate the checkout. Please keep the lease/path reserved until the real agent process teardown is complete, or add equivalent process-level cancellation proof before release.

P2: periodic disk GC now skips any persisted leased workspace path, but stale leases from dead peer processes can remain after startup. Please ensure periodic reconciliation either cleans dead persisted leases before deciding to preserve the checkout, or otherwise bounds this leak without deleting live peer leases.

majiayu000:
Blocking follow-up review finding before merge:

High: `ManagedChild` now kills the process group on drop, but the wait loop only waits the direct child. After the root CLI exits, `task_runner/spawn.rs` can release the workspace lease without proving all descendants in the killed process group have exited. That leaves a narrower version of the original cancellation race: descendant processes may still be mutating the checkout when the pool slot is reused. Please either wait for the process group to drain, use a stronger process-supervision primitive, or otherwise prove no descendants can survive before the lease is released.

Medium: stale workspace lease cleanup is still PID-only. If a dead Harness PID is reused by an unrelated process, periodic disk GC can keep treating that stale foreign lease as live. Please evaluate whether lease records should include process start time/session proof or another stronger liveness check.

majiayu000:
Blocking follow-up before merge:

The latest migration fix uses `ALTER TABLE workspace_leases ADD COLUMN IF NOT EXISTS process_started_at ...`, but SQLite rejects that syntax. Local reproduction:

```sh
sqlite3 :memory: 'CREATE TABLE t(a INTEGER); ALTER TABLE t ADD COLUMN IF NOT EXISTS b INTEGER NOT NULL DEFAULT 0;'
```

Result:

```text
Error: in prepare, near "EXISTS": syntax error
```

So `60cb29a5` fixes the duplicate-column review thread for dialects that support the syntax, but it can break SQLite-backed task DB migrations. Please replace this with a SQLite-compatible migration strategy. Two safe directions are: keep the v24 `CREATE TABLE` SQL backward-compatible and let v25 add the column, or implement an explicit column-exists check in migration code before running the `ALTER TABLE`.

majiayu000:
Resolved follow-up: the SQLite migration blocker in https://github.com/majiayu000/harness/pull/1332#issuecomment-4697165196 is addressed by the later migration split. The current PR head `6397d31a` uses v24 for the legacy `workspace_leases` shape and v25 with plain SQLite-compatible `ALTER TABLE ... ADD COLUMN process_started_at ...`; CI is green on the current head.

PR #1425, opened 2026-07-02
spec: agent run identity (cross-tool session correlation)

majiayu000:
SpecRail-style product + tech spec for a run correlation ID minted at spawn, propagated via env, bound to native session IDs in a local append-only registry. Foundation for keepline / ccstats / remem joins. Design review only — no implementation yet.


majiayu000 (commented):
Spec gate review: no blocking findings from current GraphQL reviewThreads/CI check. Proceeding to implementation follow-up.

PR #1426, opened 2026-07-02
spec: context composer (budgeted, auditable context injection)

majiayu000:
SpecRail-style product + tech spec for a single owner of harness-side context injection: provider trait, deterministic arbitration pipeline, degrade ladders, constraint-count guard, per-composition manifest, shadow mode first. CLI-hook injections (remem/vibeguard) stay untouched by design. Design review only — no implementation yet.

### DIFF
--- a/specs/context-composer/product.md
+++ b/specs/context-composer/product.md
@@ -13,7 +13,7 @@ Two failure modes, both invisible today:
 
 ## Product Behavior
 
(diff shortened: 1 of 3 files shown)


majiayu000 (commented):
Spec gate review: no blocking findings from current GraphQL reviewThreads/CI check. Proceeding to implementation follow-up.

Issue #1455, opened 2026-07-04
test: replace real sleeps with tokio virtual time in task_queue tests


majiayu000:
## Problem

Task queue tests rely on real wall-clock sleeps: 42 `tokio::time::sleep` call sites across `crates/harness-server/src/task_queue_tests.rs` and `crates/harness-workflow/src/task_queue_tests.rs` (roughly 3s of fixed waiting per file, plus flakiness from timing races). The workspace has zero uses of `#[tokio::test(start_paused = true)]`.

## Acceptance criteria

- [ ] Sleep-based tests in the two task_queue test files converted to `#[tokio::test(start_paused = true)]` with `tokio::time::advance` / auto-advance where the awaited condition is timer-driven.
- [ ] Tests that genuinely need real time (e.g. waiting on OS processes or real I/O) are left as-is with a comment stating why.
- [ ] No assertion weakening; test intent unchanged.
- [ ] `cargo test -p harness-server -p harness-workflow` passes.

## Non-goals

- DB-backed test parallelism (separate spec, see #1444).

PR #1476, opened 2026-07-04
test: replace real sleeps with tokio virtual time in task_queue tests

majiayu000:
Refs #1455

## What

Convert `harness-workflow` task_queue tests from wall-clock sleeps/timeouts to tokio virtual time via `#[tokio::test(start_paused = true)]`.

## Per-file changes

| File | Converted | Kept as-is |
|---|---|---|
| `crates/harness-workflow/src/task_queue_tests.rs` | 19 | 5 (no timers at all: `acquire_and_release_single_permit`, `running_count_and_queued_count_reset_after_completion`, `priority_zero_default_compiles`, `task_admission_blocked_under_pressure`, `task_admission_succeeds_no_pressure`) |
| `crates/harness-workflow/Cargo.toml` | +1 dev-dependency line | — |

No assertions weakened, no test removed, no intent changed — the only per-test change is the attribute. The `TaskQueue` under test uses pure tokio primitives (no `std::time::Instant`/`SystemTime`, no OS processes, no DB), so paused time is safe for every converted test; no test needed to stay real-time.

## Dependency note

`start_paused` requires tokio's `test-util` feature, added as a **dev-dependency only** of `harness-workflow` (`tokio = { workspace = true, features = ["test-util"] }`). With Cargo feature unification this affects test builds only; production builds are unaffected.

## Timing (test binary time, `cargo test -p harness-workflow task_queue`)

- Before: 24 passed, finished in **0.27s**
- After: 24 passed, finished in **0.01s**

## Verification

- `cargo test -p harness-workflow task_queue` — 24 passed, 0 failed
- `cargo fmt --all` — no diff outside the two changed files
- `RUSTFLAGS="-Dwarnings" cargo check -p harness-workflow --all-targets` — clean
- `cargo check -p harness-server` — clean (feature unification does not break the server crate)

## Follow-up

`crates/harness-server/src/task_queue_tests.rs` (the second file named by #1455) is **dead code**: it has no `mod`/`#[path]` reference anywhere in `harness-server` and is never compiled (`cargo test -p harness-server task_queue` matches 0 tests). It is a stale duplicate of the workflow copy — `harness-server` re-exports the queue via `pub use harness_workflow::task_queue` (`crates/harness-server/src/lib.rs:67`). It is intentionally left untouched here; recommend deleting it in a separate decision. Issue closure is withheld (`Refs`, not `Closes`) pending that call.

### DIFF
--- a/crates/harness-workflow/Cargo.toml
+++ b/crates/harness-workflow/Cargo.toml
@@ -24,3 +24,6 @@ tokio-stream = { workspace = true }
 async-trait = { workspace = true }
 tempfile = { workspace = true }
 sha2 = { workspace = true }
+
+[dev-dependencies]
+tokio = { workspace = true, features = ["test-util"] }
--- a/crates/harness-workflow/src/task_queue_tests.rs
+++ b/crates/harness-workflow/src/task_queue_tests.rs
@@ -22,7 +22,7 @@ async fn acquire_and_release_single_permit() {
     assert_eq!(q.running_count(), 0);
 }
 
-#[tokio::test]
+#[tokio::test(start_paused = true)]
 async fn only_max_concurrent_tasks_run_simultaneously() {
     let q = Arc::new(TaskQueue::new(&config(2, 16)));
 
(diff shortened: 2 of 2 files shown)


majiayu000:
SpecRail advisory review: no blocking findings.

Evidence checked:
- Linked issue: GH1455 (partial implementation; PR uses Refs, not Closes)
- Local verification: cargo test -p harness-workflow task_queue — 24 passed in PR worktree
- PR gate: allowed on head 5260fa5c9dfa272db02e2f8695e0fc4770a4fbaa with explicit human merge authorization

Verdict: advisory pass for this partial slice; issue closure remains separate because the stale harness-server duplicate test file was intentionally left for a later decision.

Issue #1567, opened 2026-07-05
Operator escape hatch for blocked/failed runtime workflows (unblock/retry API + structured block reason)


majiayu000:
## Problem

Once a workflow-runtime issue instance reaches `blocked`, `failed`, or `cancelled`, there is no operator-facing way to get it moving again:

- The intake coverage gate (`crates/harness-server/src/intake/github_coverage_gate.rs`, `runtime_issue_state_is_covered`) treats `blocked`/`failed`/`cancelled` as **covered**, so the poller skips the issue forever (`intake: skipping covered GitHub issue before runtime dispatch ... workflow_state=blocked`).
- No HTTP route exposes unblock/retry/resume for a workflow instance (searched `http_router.rs` — nothing matches).
- Nothing ever re-reads the GitHub issue thread for a blocked instance, so operator responses posted on the issue are never observed.

## Real-world scenario that hit this (2026-07-05, litellm-rs deployment)

litellm-rs uses a repo-level SpecRail convention with human gates. The agent correctly blocked issues #837/#841 with reasons like:

> "SpecRail task SP841-T2 requires maintainer approval before implementation; issue #841 has no approval comments"

The maintainer then posted the requested approval comments on the issues. Nothing happened — poll intake kept logging `skipping covered ... workflow_state=blocked` every tick. The only way out was manual SQL surgery (delete the instance + its commands/jobs/events from `workflow_runtime.*`) followed by re-dispatch. That is not an acceptable operator workflow, and editing workflow tables directly is explicitly discouraged by the runtime's own contract.

This is a general gap, not a SpecRail-specific one: any agent-initiated `blocked` (waiting for human input) or transient `failed` (e.g. the 173-job codex proxy outage the same day) currently requires DB surgery to recover.

## Proposed direction (minimal, policy stays in-repo)

1. **Structured block reason**: when an activity result transitions an instance to `blocked`, persist a machine-readable reason field (`blocked_reason`, plus optional `unblock_hint`, e.g. "await approval comment on issue #841") instead of only free-text inside event JSON.
2. **Operator API**: `POST /api/workflows/instances/{id}/unblock` (and `/retry` for `failed`) that resets the instance to a re-dispatchable state and clears the coverage hold. Auth via the existing bearer token.
[...]

PR #1579, opened 2026-07-10
Add workflow runtime unblock and retry routes

majiayu000:
## Summary

- Add authenticated workflow runtime `/api/workflows/runtime/unblock` and `/api/workflows/runtime/retry` routes for GitHub issue PR workflows.
- Persist operator recovery audit metadata while preserving structured stop metadata, validate recovery transitions, and re-dispatch stopped issue workflows without direct workflow table edits.
- Constrain stopped-state recovery transitions to `workflow_runtime_operator_action` plus `operator_runtime_unblock` / `operator_runtime_retry`, preserving the human merge gate instead of making `blocked -> merging` generally valid.
- Replay the exact original stopped command through `last_stop.runtime_job_id` for structured stops so phase-specific command type and payload are preserved.
- Support production-shaped `last_stop.activity = "start_child_workflow"` for PR-feedback child workflow replay, and require nonblank `subject_key` with the existing `definition_id`, child activity, and PR-number validation.
- Add a safe legacy fallback for GitHub issue workflows only when structured stop metadata is truly absent/null and no structured `error_kind` is present: failed retry and blocked unblock move only to the generic `implement_issue` operator recovery path, preserving existing evidence and operator context.
- Fail closed for present non-null `last_stop` objects that are empty, partial, event-id-only, all-null, or otherwise lack a usable replay path; unblock/retry return unsupported stopped activity instead of silently dispatching `implement_issue`.
- Treat malformed `last_stop` typed fields (`state`, `activity`, `runtime_job_id`, `error_kind`) and malformed non-object `last_stop` values as recovery errors on unblock/retry instead of legacy/absent metadata.
- Supersede stale active commands during operator recovery: skip pending/dispatching commands, cancel unfinished jobs for stale dispatched commands, and keep the existing lock order so command-row waits do not hold the workflow instance lock.
- Retain compact route coverage for success, not found, store unavailable, wrong-state, non-retryable, and partial-stop-metadata 409 responses.

Related: #1567

Refs #1567

## Validation

- `cargo fmt --all -- --check` passed
- `cargo test -p harness-workflow runtime_recovery -- --nocapture` passed
- `cargo test -p harness-server workflow_runtime_recovery_endpoints_cover_contract -- --nocapture` passed
- `RUSTFLAGS="-Dwarnings" cargo check --workspace --all-targets` passed
- `cargo clippy --workspace --all-targets -- -D warnings` passed
- `git diff --check` passed
- Explicit scope guard passed against base `d00ade2a9c587f6c79d1347637e8e181ecfdec90`: 12 files changed, 1500 added lines, 6 deleted lines; at or below max 30 files and 1500 added lines.

## Review

- Addressed active blocking feedback on recovery-only transitions, phase-specific command replay, legacy failed-row retry compatibility, legacy blocked-row unblock compatibility, lock ordering, replay matcher strictness, endpoint-level evidence, auth middleware coverage, malformed stop metadata, stale PR body/scope reporting, and present-but-incomplete stop metadata fallback safety.
- Recovery tests assert replayed command type and payload, including pr-hygiene `address_pr_feedback` fields (`source`, `review_summary`, `hygiene`), a generic `address_pr_feedback` replay without PR payload, production-shaped `start_child_workflow` metadata, missing child `subject_key`, malformed stop metadata on retry and unblock, direct malformed `last_stop: 42`, legacy blocked unblock, absent/null legacy fallback, rejection of empty/event-id-only/all-null `last_stop` objects, no mutation for rejected partial metadata, and audit-trail event/data persistence.
- Endpoint tests retain unblock/retry success coverage plus not found, store unavailable, wrong-state, non-retryable, and partial-stop-metadata response mappings.
- No review threads were marked resolved, no merge was requested, and issue #1567 remains open.

## Out of Scope

- Intake coverage behavior for untouched stopped workflows, operator monitor/runtime tree exposure, web UI actions, usage-guide docs, and optional blocked recheck policy remain for T004-T008.


majiayu000:
Independent SpecRail review found one blocking lifecycle issue: operator recovery currently always transitions to `implementing` and enqueues `implement_issue`, even when `last_stop.activity` identifies a stopped `merge_pr`, `run_local_review`, `sweep_pr_feedback`, `address_pr_feedback`, or `replan_issue` activity. That can report recovery success while duplicating implementation or skipping the failed lifecycle phase. Please derive and validate the supported recovery state/activity from structured stop metadata, or explicitly reject unsupported stopped activities instead of silently dispatching `implement_issue`. Add regression coverage for non-implementation stopped activities. This must be resolved before T002/T003 are considered complete. No merge requested.

crates/harness-workflow/src/runtime/store/recovery.rs
```diff
@@ -0,0 +1,426 @@
...
+    WorkflowCommand::new(
+        WorkflowCommandType::EnqueueActivity,
+        format!(
+            "operator-recovery:{}:{}:{}",
+            action.as_str(),
+            instance.id,
+            event_id
+        ),
+        json!({
+            "activity": "implement_issue",
```
majiayu000: Blocking: recovery_dispatch_command unconditionally enqueues implement_issue, but blocked or failed workflows can originate from merge_pr, run_local_review, sweep_pr_feedback, address_pr_feedback, replan_issue, and other activities. Unblock/retry therefore reports success while replaying the wrong phase, risking duplicate implementation or lifecycle skips. Derive and validate the recovery activity and next state from persisted stop metadata, or explicitly reject unsupported activities; add phase-specific tests including at least one non-implementation stop.

majiayu000 (commented):
Blocking recovery-semantics review.

crates/harness-workflow/src/runtime/store/recovery.rs
majiayu000: Additional blocking check: the current PR worktree is 11 files with 1,602 additions relative to origin/main, exceeding WORKFLOW.md pr_scope_guard.max_lines_added=1,500 by 102 lines. Compact the implementation/tests to at most 1,500 additions and run the repository pr_scope_guard before pushing or reporting this thread resolved.

majiayu000: Submission constraint: the repair commit is already on the remote branch. Preserve history per the repository L7 rule—create a follow-up compaction commit and use a normal push. Do not amend the pushed commit and do not force-push.

majiayu000:
Addressed the recovery-semantics feedback in head 831919af. Recovery now derives the resumed workflow state/activity from last_stop.activity for implement_issue, replan_issue, merge_pr, run_local_review, sweep/inspect PR feedback, and address_pr_feedback; unsupported stopped activities return a conflict instead of silently enqueueing implement_issue. Added regression coverage for non-implementation stopped activities.

Scope guard now passes at 11 files / 1497 additions against origin/main. Validation run locally: cargo fmt --all -- --check; cargo test -p harness-workflow runtime_recovery -- --nocapture; cargo test -p harness-server workflow_runtime_recovery -- --nocapture; cargo test -p harness-server workflow_runtime_unblock_endpoint_reopens_blocked_workflow -- --nocapture; cargo test -p harness-server workflow_runtime_retry_endpoint_reopens_retryable_failed_workflow -- --nocapture; cargo clippy --workspace --all-targets -- -D warnings.

History note: I initially compacted by amend/force-push before seeing the later history-preservation comment. I repaired that with a normal pushed merge commit (831919af) that makes the previously pushed repair commit f2eda33d reachable again while preserving the compacted tree. No review threads were marked resolved.

crates/harness-workflow/src/runtime/validator.rs
```diff
@@ -131,6 +131,33 @@ impl TransitionAllowlist {
...
+                "blocked",
+                "awaiting_feedback",
+                [EnqueueActivity, StartChildWorkflow, Wait],
+            )
+            .allow(
+                "blocked",
+                "addressing_feedback",
+                [EnqueueActivity, StartChildWorkflow, Wait],
+            )
+            .allow("blocked", "merging", [EnqueueActivity])
```
majiayu000, line 160: Blocking human-gate bypass: these new blocked-state transitions live in the shared allowlist. Because blocked is nonterminal, an ordinary runtime decision can now validate blocked -> merging with EnqueueActivity; the GitHub-specific validator does not require the operator recovery actor or decision name. Constrain every newly added stopped-state recovery transition—especially blocked -> merging—to context.actor == workflow_runtime_operator_action plus operator_runtime_unblock/operator_runtime_retry, and add a negative test proving an ordinary runtime actor cannot validate it.

crates/harness-workflow/src/runtime/store/recovery.rs
```diff
@@ -0,0 +1,528 @@
...
+            "operator-recovery:{}:{}:{}",
+            action.as_str(),
+            instance.id,
+            event_id
+        ),
+        payload,
+    )
+}
+
+const RECOVERY_CONTEXT_FIELDS: &[&str] = &[
```
majiayu000: Blocking recovery correctness: this reconstructed command copies only common workflow fields. A pr_hygiene address_pr_feedback command requires source=pr_hygiene, review_summary, and hygiene, while the workflow persists hygiene_context; retry currently drops those instructions and silently changes the operation. Replay the original stopped command using last_stop.runtime_job_id, reconstruct and validate all activity-specific fields, or reject context-dependent activities. Tests must assert payload/command type, not only state and activity name.

majiayu000 (commented):
Request changes. The original always-implement_issue bug is fixed, but the current patch opens a blocked-to-merge human-gate bypass and loses phase-specific command context. Also refresh the stale PR body after the code is corrected: it still describes unconditional implement_issue recovery and the old 1,258-line scope.

crates/harness-workflow/src/runtime/store/recovery.rs
```diff
@@ -0,0 +1,528 @@
...
+                workflow: instance.clone(),
+                error_kind,
+            });
+        }
+    }
+
+    None
+}
+
+fn recovery_dispatch_target(data: &Value) -> Result<RecoveryDispatchTarget, Option<String>> {
```
majiayu000: Blocking spec gap: GH1567 tech.md requires legacy failed workflows with missing error_kind to remain retryable. Pre-T001 rows also lack structured last_stop.activity, but recovery_dispatch_target rejects them, so the operator API cannot recover the legacy case the spec explicitly preserves. Add a safe legacy fallback (for example an earliest safe redispatch path) with a regression test, while continuing to reject present-day unknown/context-dependent stopped activities.

majiayu000 (commented):
Additional blocking SpecRail finding from Harness local review.

majiayu000:
Implementation guard for the current repair round: (1) database/query/deserialization failures while loading the stopped command must propagate as internal errors; do not convert them into UnsupportedStoppedActivity, per U-17/U-29. (2) validate the replayed command type and phase-specific payload against the recovery target. In particular, do not accept every non-EnqueueActivity command automatically; a StartChildWorkflow replay must have the expected child activity/context. Add or preserve focused negative coverage.

majiayu000:
Scope-compaction guard: do not delete all endpoint-level evidence for the GH1567 acceptance criteria. T002/T003 explicitly require authenticated routes plus wrong-state and non-retryable rejection. Keep a compact route test that exercises at least wrong-state and non-retryable 409 mapping, and retain/cite concrete middleware auth coverage for these routes. Reduce duplication or production boilerplate instead of removing the only behavior assertions.

majiayu000:
Blocking lock-order regression in the current compaction: do not lock workflow_instances before workflow_commands. Runtime completion/dispatch paths take command rows first, and runtime_recovery_waiting_on_command_does_not_lock_instance exists to prevent the command↔instance deadlock. Keep commands -> instance ordering. A compact safe shape is: supersede/lock commands inside the transaction, then select_instance_for_update and validate the state/dispatch plan; any wrong/unsupported/not-found outcome rolls the transaction back, so the provisional command updates do not persist.

majiayu000:
Replay matcher correction: do not require pr_number on every address_pr_feedback or run_local_review command. Existing legitimate paths use WorkflowCommand::enqueue_activity(...) and therefore carry only activity (for example build_pr_feedback_decision and the post-rework local-review enqueue). Requiring pr_number would turn recoverable current stops into UnsupportedStoppedActivity. Exact activity matching is sufficient for generic enqueue commands because workflow data supplies bound PR context; require pr_number/review_summary/hygiene only for the explicit source=pr_hygiene variant. Add/retain a minimal-command regression or keep the matcher narrow.

majiayu000:
Blocking verification note for the current rework: scope compaction must retain real positive HTTP integration coverage for both POST /api/workflows/runtime/unblock and POST /api/workflows/runtime/retry. Store-layer success tests alone do not prove the T002/T003 route contract. Keep status/body plus resulting workflow/replay evidence for both success paths, along with wrong-state, nonretryable, and query-token auth cases. Compact helpers or other redundancy instead of deleting these interface acceptance assertions.

majiayu000:
Additional spec check: GH1567 tech.md lines 221-226 explicitly require HTTP route tests for success, not found, store unavailable, wrong state, and non-retryable failure responses. The final PR gate must retain evidence for all five response classes; query-token auth rejection remains a separate security assertion. A direct unit test of runtime_recovery_response may cover response mapping, but it cannot replace real handler/router coverage for authenticated success and store-unavailable behavior.

crates/harness-workflow/src/runtime/store/recovery.rs
majiayu000: Addressed at final head 4a5c4943177e563c2ad69e8c5e26cabaf1e73e7c. Recovery now validates last_stop/runtime_job ownership and replays the exact supported source command instead of unconditionally dispatching implement_issue; unsupported activities fail closed. The final scope is exactly 12 files and 1,500 additions, and pushed history was preserved through normal follow-up commits.

crates/harness-workflow/src/runtime/validator.rs
majiayu000: Addressed at final head 4a5c4943177e563c2ad69e8c5e26cabaf1e73e7c. validator_github_issue_pr.rs now requires actor workflow_runtime_operator_action plus decision operator_runtime_unblock/operator_runtime_retry for every stopped-state recovery transition. The negative regression rejects an ordinary runtime actor attempting blocked -> merging. Fresh Gemini review on this exact head reported no comments.

crates/harness-workflow/src/runtime/store/recovery.rs
majiayu000: Addressed at final head 4a5c4943177e563c2ad69e8c5e26cabaf1e73e7c. Recovery replays the persisted source command and validates activity-specific payloads, including pr_hygiene source/review_summary/hygiene data and start-child workflow context. Regression tests assert command type and payload.

crates/harness-workflow/src/runtime/store/recovery.rs
majiayu000: Addressed at final head 4a5c4943177e563c2ad69e8c5e26cabaf1e73e7c. The legacy fallback is limited to truly absent/null last_stop with no structured error_kind; present partial, malformed, or unsupported stop metadata fails closed. Regression tests cover legacy failed retry and malformed/partial metadata rejection.

PR #1581, opened 2026-07-11
feat(runtime): operator unblock/retry escape hatch (GH-1567 T002-T004)

### LINKED ISSUE #1567: Operator escape hatch for blocked/failed runtime workflows (unblock/retry API + structured block reason)

majiayu000:
Closes part of #1567 (T002-T004; does not fully close the issue).

## Problem

Blocked and failed workflow-runtime instances get stuck forever: the coverage gate treats `blocked`/`failed` as covered, so intake skips the issue and the only recovery was direct Postgres surgery (the exact pain #1567 removes).

## Change

Operator recovery actions that move a stopped `github_issue_pr` instance back into a dispatchable state and re-enqueue the plan activity, clearing the coverage hold on the next intake tick:

- `POST /api/workflows/runtime/unblock` — blocked-only
- `POST /api/workflows/runtime/retry` — failed-only; rejects non-retryable `fatal`/`configuration`

Both are body-based (`workflow_id` + optional `reason`) to match the existing runtime merge/cancel routes and avoid slash-encoding of workflow ids, run behind the existing auth middleware, write an audit event, and preserve historical `last_stop` evidence while clearing the active stop-reason fields.

- validator: allow `blocked → {scheduled,planning,implementing}`, mirroring the existing `failed →` recovery transitions
- coverage gate unchanged — T004 is satisfied by the state transition: `blocked`/`failed` stay covered; recovery moves state to `planning` (uncovered), so the next tick re-dispatches

## Scope

- T001 (structured stop metadata) landed in #1573.
- This PR: T002 / T003 / T004.
- Supports `github_issue_pr` (the litellm-rs scenario); prompt_task recovery is a follow-up.
- T005-T008 (observe / web / docs / policy) remain follow-ups. Do not close #1567 on merge.

## Verification (local, this session)

- `cargo test -p harness-server recover_tests` → 5 passed
- validator suite → 11 passed; `workflow_runtime_submission` regression → 38 passed
- `cargo check --workspace --all-targets`, `cargo clippy -p harness-server -p harness-workflow --all-targets -- -D warnings`, `cargo fmt --all --check` → clean


majiayu000:
Superseded by #1579, which merged the same GH-1567 T002-T003 operator unblock/retry routes first — and with a stronger approach: it resumes the actual stopped activity (`last_stop_activity`) with richer `blocked → {implementing,replanning,local_review_gate,merging}` transitions, versus this PR's simpler "always re-plan to planning". Closing as a duplicate to avoid conflicting work.

No unique value remains in this branch: T004's coverage-gate behavior is satisfied by the state transition in #1579 as well (blocked/failed stay covered; recovery moves state off the stopped state). Remaining #1567 follow-ups (T004 marking, T005-T008) are tracked separately.

Issue #514, opened 2026-03-23
[P0] REVIEW-02: read-only review turns lost hard tool isolation


majiayu000:
The recent review-path changes removed hard tool allowlists from both agent CLIs, but the replacement validator only scans final stdout text for tool names.

Why this matters:
- `crates/harness-agents/src/claude.rs:69` explicitly stops passing `--allowedTools` even though review turns still run with `--dangerously-skip-permissions`.
- `crates/harness-agents/src/codex.rs:58` never applies `req.allowed_tools` to Codex CLI arguments either.
- `crates/harness-server/src/task_executor.rs:733`, `:994`, and `:1176` now rely on `validate_tool_usage(...)` after execution.
- `crates/harness-server/src/task_executor.rs:292-324` reconstructs `AgentResponse.output` only from `MessageDelta` / `AgentReasoning`, and `crates/harness-agents/src/streaming.rs:161-169` emits plain stdout lines, not authoritative tool-call events.

In practice this makes read-only review/check turns fail open: a model can call `Write`, `Edit`, or mutating `Bash` commands, and harness will miss it unless the model literally prints `<Write>` or JSON tool metadata in its final text output.

Recommended action:
1. Restore hard tool enforcement at the adapter/CLI boundary for restricted turns, or switch review execution onto an adapter path that emits structured tool-call events.
2. Treat tool policy violations as pre-dispatch / in-band enforcement, not a best-effort regex over final output.
3. Add an integration test proving a review-mode task cannot mutate files when only read tools are allowed.

PR #1726, opened 2026-07-24
fix(agents): apply sandbox spec and align CLI args in ClaudeAdapter

majiayu000:
ClaudeAdapter::start_turn diverged from ClaudeCodeAgent (claude.rs):

- Never built a SandboxSpec, so capability-token allowed_write_paths
  were silently dropped on the adapter path; now routed through
  spawn_contract::prepare_agent_spawn like claude.rs and codex_adapter.
- Ignored req.env_vars entirely; now propagated via the spawn contract.
- Ignored req.sandbox_mode; now honored (DangerFullAccess default,
  matching codex_adapter).
- Passed --dangerously-skip-permissions together with --allowedTools,
  which are mutually exclusive in Claude CLI 2.1.70+; now uses the
  issue-#514 split (--permission-mode bypassPermissions + --allowedTools
  for restricted profiles).

Test module extracted to claude_adapter_tests.rs (file exceeded the
800-line ceiling) with three new tests: full-profile flag, restricted
profile mutual exclusion, env-var propagation.


majiayu000:
Addressed: TurnRequest.allowed_tools is now Option<Vec<String>> (None = full profile, Some(empty) = deny-all, never promoted to --dangerously-skip-permissions), with a shared uses_dangerously_skip_permissions() helper and a new deny-all regression test. The redundant AsRef conversion is also removed.

PR #1727, opened 2026-07-24
fix(server): fail closed on malformed WORKFLOW.md instead of silent defaults

majiayu000:
Two call sites swallowed load_workflow_config errors with
unwrap_or_default(), while every other call site propagates them:

- http/builders/registry.rs: a malformed project WORKFLOW.md silently
  booted the server with the default storage.schema_namespace and
  default workflow policy, with no log line. Now aborts startup with
  context. A missing WORKFLOW.md still resolves to the base/default
  config without error.
- workflow_runtime_worker/data_helpers.rs
  force_execute_from_project_policy: a malformed WORKFLOW.md silently
  reverted to the default force-execute label. Now returns an error the
  runtime job propagates.

Tests: malformed front matter errors with load-failure context; missing
WORKFLOW.md still matches the default force-execute label.

### DIFF
--- a/crates/harness-server/src/http/builders/registry.rs
+++ b/crates/harness-server/src/http/builders/registry.rs
@@ -1,3 +1,4 @@
+use anyhow::Context;
 use dashmap::DashMap;
 use std::path::Path;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


majiayu000:
Addressed: force_execute_from_project_policy now runs under tokio::task::spawn_blocking in the async start_child_workflow path, with join errors propagated (no expect).

Issue #1434, opened 2026-07-03
refactor(kernel): Phase 1 — remove zero-traffic layers (thread/turn RPC, task layer, harness-eval, review_store)


majiayu000:
## Problem

Kernel-contraction Phase 1 per the architecture analysis (better/reports/harness-kernel-design.md, decisions approved 2026-07-03). Production data shows severe volume/value inversion:

- workflow runtime: 216,161 runtime_events, 67,436 jobs, 1,877 workflows — carries all real traffic
- thread/turn lifecycle layer (16 JSON-RPC methods, thread_manager/thread_db): **9 threads ever**
- task layer (task_db/task_executor/task_runner): **62 tasks ever**
- harness-eval (848 LOC): **0 eval_runs**
- review_store: **0 review_findings** (review actually happens via GitHub bots)

These are three generations of state machines stacked on top of each other; the third (workflow runtime) has factually replaced the first two. Direct delete was chosen over plugin extraction (git history is the archive).

## Search Evidence

- psql aggregations over event_store / workflow_runtime / task_db / thread_db / eval_store / review_store (2026-07-03 session, reproduced in the design doc).
- LOC census: harness-server 100,538 of 158k total; target ≤40k after full contraction.
- Design principle: capability moves into models; authority (policy), evidence (verify/audit), and resource control stay external. GitHub issues/PRs/checks are the durable state machine — no self-built thread/turn/task layers.

## Desired Outcome

1. **Safety probe first**: usage counters on every surface slated for deletion; 7-day zero-count gate (or explicit maintainer waiver) before each removal lands.
2. **Data archived, then code deleted**: pg_dump the 9 threads + 62 tasks + related tables to a dated archive; tables kept (renamed/archived), never dropped in this phase.
3. Delete: thread/* and turn/* RPC methods (16), thread_manager/thread_db, task_db/task_executor/task_runner (merge whatever workflow runtime still needs), harness-eval crate, review_store module + schema usage.
4. All consumers cleaned: CLI subcommands, dashboard panels, router registrations, protocol definitions referencing the deleted surfaces.
5. `cargo test --workspace` green; `harness pr fix`, `harness status`, webhook intake, and the dashboard behave unchanged for workflow-runtime paths.

## Non-Goals

- Phase 2 (verify-over-envelope) and Phase 3 (intelligence-layer removal: rules/skills/GC/learn/self_evolution/q_value/complexity_router) — separate issues.
[...]

PR #1729, opened 2026-07-24
refactor: dedupe SignalThresholds, drop unused CLI dep, refresh U-16 table

majiayu000:
- Delete harness-gc's SignalThresholds struct (field-for-field duplicate
  of harness_core::config::misc::SignalThresholdsConfig with identical
  defaults) plus its From bridge, the CLI's hand-written map_thresholds
  mirror, and the identity map_gc_config wrapper. SignalDetector now
  takes SignalThresholdsConfig directly.
- Remove the unused harness-protocol dependency from harness-cli
  (zero references in src/).
- Refresh the CLAUDE.md U-16 exemption table: six entries pointed at
  files already split (or deleted) far below their exempted ceilings —
  services/execution.rs 2600→537 actual, task_runner/store.rs 2200→82,
  workflow runtime store.rs 2300→290, webhook.rs 1000→335, etc. Add the
  actually-oversized prompts/parsing.rs, tighten misc_routes.rs, and
  mark spawn.rs/triage_pipeline.rs as pending GH-1434 deletion.

Tests: harness-gc 81 pass, harness-cli 43 pass, workspace clippy clean.


majiayu000:
Addressed: switched to Default::default() at the suggested sites.

Issue #1771, opened 2026-07-26
Prompt-injection surface: permissive default agent profile, no first-party egress enforcement


majiayu000:
**Narrowed 2026-08-06:** direction (1) — taint-tracking provenance of `workflow.data` and fencing agent-written values in prompt packets — landed in #1851 (`825a5903`). The two remaining directions below are the open scope of this issue.

## Problem

Defenses exist but the trust boundary is inconsistent:

- `wrap_external_data()` fences issue bodies, `repo_memory` carries an untrusted preamble, and prompt packets fence tainted `workflow.data` (#1851) — good.
- The default Claude profile runs `--dangerously-skip-permissions` (`claude.rs:131`); the scoped-tools profile exists but is not the default.
- Egress allowlisting is not enforced by harness — it is exported as `HARNESS_AGENT_EGRESS_ALLOWLIST` for an external proxy that is not part of the deployment; without `HARNESS_AGENT_EGRESS_PROXY` the container tier falls back to `--network none` and the host tier has no egress control.

## Proposed direction

(2) Scoped permissions as the default profile with an explicit opt-up to full; (3) first-party egress enforcement for the host tier or a bundled proxy for the container tier.

Full analysis: `docs/references/prompt-injection-hardening.md`.

majiayu000:
## Remediation plan (verified against origin/main `1a6f475b` + PR #1851 head, 2026-08-04)

Status check first: of this issue's three proposed directions, PR #1851 implements **direction (1)** — field-level `workflow.data` provenance sidecar (`harness-workflow/src/runtime/data_provenance.rs`), v3 packet with fenced untrusted workflow data + continuation context, and (beyond the issue text) `command_input` taint splitting. Directions **(2) scoped default profile** and **(3) egress enforcement** are untouched by any open PR. Plan below.

### Phase 0 — land PR #1851 (blocker for everything in direction 1)

The PR is currently `CONFLICTING` with main and its `Test` CI job fails. Concrete finish list is on the PR thread; the P1s that matter for this issue's threat model:
- No production code path ever assigns `data_provenance` to an instance, so every real workflow stays in legacy mode forever (fence-everything + degradation artifact — safe, but the classification feature is inert).
- Descendant provenance overrides are not honored before trusting a container (`/snapshot` Server + `/snapshot/body` External renders the whole object trusted).
- `project.repo` is projected from workflow data into the trusted packet section outside the fence (`prompt_packet.rs:59`).
[...]

PR #1779, opened 2026-07-26
spec(GH1771): define prompt injection hardening

majiayu000:
Defines the fail-closed trust-boundary contract for workflow-runtime prompts, with a full analysis report under docs/references/.

- **Analysis** (`docs/references/prompt-injection-hardening.md`): maps existing defenses (external-data fencing, repo-memory untrusted preamble, sandbox write-deny) against the gaps — agent-written `workflow.data` replayed as trusted state across turns, `--dangerously-skip-permissions` as the default profile in both spawn paths, and egress control delegated to an unshipped external proxy (host tier has none).
- **Spec** (`specs/GH1771/`): (1) field-level provenance on `workflow.data` (server/agent/external) with agent- and external-origin values re-entering prompts only inside untrusted fences, fail-closed on unclassified fields under packet schema v2; (2) scoped tool profile as the runtime default with explicit recorded opt-up to the full profile; (3) egress enforcement contract — host-tier deny-by-default allowlist, bundled-proxy option for the container tier, preserved `--network none` fallback.

Complements GH1732 (provenance recording) with enforcement; consumes its trust taxonomy rather than replacing the manifest.

Refs #1771


majiayu000:
**Independent agent review: APPROVE** (after 3 rounds, fail-closed gate)

- Round 1: REQUEST_CHANGES — a defense-table row cited a symbol deleted with the legacy layer (#1725), and the v2 fencing gate conflicted with the merged GH1732 schema mandate. Fixed in 2e3ad83e (real sandbox mechanism cited; v3 packet schema with migrated_at grandfathering + fail-closed typed error).
- Round 2: caught a new misstatement in the replacement text ('Claude spawn paths take no sandbox mode') — wrong: runtime profile sandbox mode flows agent-agnostically into Claude spawns as an OS-level Seatbelt/Landlock sandbox; only the tool-permission flag surface is sandbox-unaware. Corrected in 979fc51a.
- Round 3 targeted verification: PASS — every cited chain link re-verified against origin/main.

Review performed by fresh-context agent processes that did not author the changes.

Issue #1772, opened 2026-07-26
Runtime reliability: dual leases can disagree, completed work is silently dropped, retention/watchdog off by default, unsupervised background loops


majiayu000:
## Problem

- **Dual lease systems**: runtime jobs lease in \`runtime_jobs\` (owner + \`lease_generation\`, TTL 600s); workspaces lease in \`workspace_leases\` (owner_session + PID liveness). Nothing ties them atomically. A worker whose job lease expires mid-turn has its completion rejected — \`commit_runtime_activity_completion_with_transcript_if_owned\` returns \`None\` and the worker \`tracing::warn!\`s and drops the entire result — while the agent process and workspace lease keep running. Work lost, workspace occupied, second worker re-runs against a dirty tree.
- **\`WorkflowInstance.version\` is a decoy**: serialized and named like a CAS token, but every mutation relies on \`SELECT ... FOR UPDATE\` + string \`expected_state\`; any future writer using \`version\` optimistically will silently lose updates.
- **Retention off by default**: \`runtime_retention_enabled: false\`, \`task_retention_enabled: false\`, \`workflow_watchdog_enabled: false\` — events, decisions, and full transcripts grow without bound, and stuck \`blocked\`/\`awaiting_feedback\` workflows are never surfaced (the orphan-schema reaper's own doc comment records this declaration-execution-gap class already biting once).
- **Single-process orchestration**: ~10 bare \`tokio::spawn\` loops from \`http/mod.rs\` with warn-and-continue failure handling, no leader election, and \`WORKFLOW.md\` re-parsed every tick — a malformed config file silently disables retention, watchdog, and reaping simultaneously while dispatch keeps running.

## Proposed direction

(1) Bind job lease renewal to workspace lease (single ownership epoch); route lease-expired completions to a dead-letter/reconcile path instead of dropping; (2) either enforce version-CAS or remove/rename the field; (3) ship safe retention defaults; (4) supervised background tasks with health surfacing, and config-parse failure treated as loud degraded mode.

Full analysis: \`docs/references/runtime-reliability.md\` (spec PR follows).

majiayu000:
Supplementary evidence from an independent architecture review of origin/main @ 81c78255, supporting directions (2) and (4):

[...]

PR #1780, opened 2026-07-26
spec(GH1772): define runtime reliability hardening

majiayu000:
Defines the fail-closed behavior contract for four runtime reliability defect clusters: dead-letter persistence for lease-rejected activity completions plus an ownership epoch binding runtime job leases to workspace leases; resolution of the unchecked `WorkflowInstance.version` field (enforce-as-CAS or rename); safe terminal-only retention defaults with an operator-visible dry-run; and supervised background loops with a single loud degraded mode on `WORKFLOW.md` parse failure (including dispatch pause after a grace period).

Adds `docs/references/runtime-reliability.md` — the full evidence-based analysis (all file:line citations verified on main @ 81c78255) — and the spec packet `specs/GH1772/{product,tech}.md`.

Refs #1772

### DIFF
--- a/docs/references/runtime-reliability.md
+++ b/docs/references/runtime-reliability.md
@@ -1,0 +1,284 @@
+# Runtime Reliability Analysis: Leases, Versioning, Retention, and Loop Supervision
+
+> Status: analysis reference for GH-1772
+> Date: 2026-07-26
+> Method: read-only source inspection on `main` (81c78255); all file:line
+> citations verified against that revision.
+
+## Scope
+
+This report analyzes four reliability defect clusters in the workflow runtime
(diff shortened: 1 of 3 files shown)


majiayu000:
**Independent agent review: APPROVE** (16 citations checked)

Minor findings addressed in 5e527078: store-function file attributions corrected (runtime_job_state.rs / activity_completion.rs), the version-predicate grep claim narrowed to workflow_instances with the definitions-table counterexample cited, and the D4 advisory-lock paragraph demoted to an explicitly non-normative note to match product Non-Goals.

Review performed by a fresh-context agent process that did not author the changes.

Issue #1792, opened 2026-07-26
refactor(agents): collapse CodeAgent/AgentAdapter into one AgentBackend trait


majiayu000:
## Problem

Two parallel agent traits live in `harness-core/src/agent.rs`: `CodeAgent` (`:15`, execute/execute_stream over `AgentRequest`/`StreamItem`) and `AgentAdapter` (`:358`, start_turn/interrupt/steer over `TurnRequest`/`AgentEvent`). `TurnRequest` and `AgentRequest` are ~90% the same struct; `StreamItem` and `AgentEvent` are overlapping event enums bridged **twice** (`claude_stream.rs:67-122` inside the agents crate, `turn_lifecycle.rs:15-59` again in the server, plus an `emitted_agent_completion` dedup dance that exists only because two enums disagree about who owns the final message).

Net effect: adding a new CLI backend (e.g. gemini) touches 7+ files across 3 crates, and `AdapterExecutionStrategy` exists only to disambiguate the two traits.

## Proposed direction

One trait:

```
trait AgentBackend {
  fn name(&self) -> &str;
  fn capabilities(&self) -> Capabilities;          // steer/interrupt/approve/stream flags
  fn plan(&self, req: &TurnRequest) -> Result<SpawnPlan>;   // argv + env only (pure)
  fn parse_line(&self, line: &str) -> LineOutcome;          // stdout -> events
  async fn control(&self, op: ControlOp) -> Result<()>;     // default Unsupported
}
```

`plan()` already exists as `base_args` in claude.rs/codex.rs (pure, unit-tested); `parse_line()` as the two stream parsers; execution belongs to the shared supervisor (see the spawn-pipeline issue). Codex's stateful app-server session stays a per-turn backend instance via the existing factory registration. Delete `AgentRequest` or `TurnRequest`, one event enum, and `AdapterExecutionStrategy`.

## Acceptance criteria

- One request type, one event enum, zero bridging conversions.
- New-backend checklist is: argv builder + line parser + registry entry.
- Tolerant-vs-strict line-parse policy (Claude swallows, Codex hard-errors today) becomes one shared decision.

Evidence from static architecture review of origin/main @ 81c78255; line numbers may drift.

PR #1807, opened 2026-07-26
fix(GH1783): honor Claude result is_error and surface tool_use blocks

### LINKED ISSUE #1783: fix(agents): Claude stream-json result parsing ignores is_error/subtype and never surfaces tool_use
## Problem

Two defects in Claude CLI stream-json parsing:

1. **Failed turns are reported as success.** The `"result"` arm reads only `.result` and ignores `is_error` / `subtype`. A Claude CLI failure result (`{"type":"result","subtype":"error_during_execution","is_error":true,...}`) emitted with exit code 0 is mapped to a completed turn with empty output. `grep` confirms neither `is_error` nor `error_during_execution` appears anywhere in `crates/harness-agents/src`.
2. **Tool calls are never surfaced.** `parse_stream_json_line` matches top-level `"tool_use"` / `"tool_result"` message types, but the CLI emits tool use as content blocks inside `{"type":"assistant","message":{"content":[...]}}`; `parse_assistant_text` filters to `"text"` blocks only. Those two match arms are dead code, and the Claude backend produces a strictly poorer trajectory than the Codex backend (which does parse item events) — an invisible asymmetry at the `StreamItem` boundary.

## Evidence (origin/main @ 81c782555)

- `crates/harness-agents/src/claude_adapter.rs:369-402` (`parse_stream_json_line`), `:384-391` (result arm), `:404-423` (`parse_assistant_text`).
- `crates/harness-agents/src/codex_exec_parser.rs:122-142` (Codex surfaces tool items).

## Proposed direction

Honour `is_error`/`subtype` on `result` (map to a failed-turn event carrying the subtype); parse `tool_use` content blocks from assistant messages into tool events; add fixture tests from recorded CLI transcripts.

## Acceptance criteria

- A recorded failure-result fixture produces a failed turn, not an empty success.
- Tool calls appear in the Claude backend's event stream with name/input.
- Backend parity test: equivalent transcripts yield comparable trajectories for Claude and Codex.

majiayu000:
Closes #1783.

The Claude CLI emits terminal failures as `{"type":"result","is_error":true,"subtype":"error_*"}` with exit code 0; both `execute()` and `execute_stream()` treated these as success with empty output. Tool calls never surfaced because the parser matched a top-level `tool_use` shape the CLI does not emit, while real `tool_use` content blocks in assistant messages were dropped.

- `parse_stream_json_events` parses every event on a line in content-block order (adjacent text merged into one delta, preserving output shape)
- error results become `Error` events and fail the turn on both execution paths; usage from failed results is still recorded
- fixture tests for failure results (exit 0) and tool_use blocks

Verification: `cargo test -p harness-agents` (209 passed), `cargo clippy --workspace --all-targets -- -D warnings` clean.

### DIFF
--- a/crates/harness-agents/src/claude.rs
+++ b/crates/harness-agents/src/claude.rs
@@ -285,6 +285,12 @@ impl CodeAgent for ClaudeCodeAgent {
             )));
         }
 
+        // The CLI reports terminal failures via a `result` event while still
+        // exiting 0, so the exit-status check above is not sufficient.
+        if let Some(failure) = parsed.failure {
+            return Err(harness_core::error::HarnessError::AgentExecution(failure));
+        }
+
         Ok(AgentResponse {
             output: parsed.output,
             stderr,
--- a/crates/harness-agents/src/claude_adapter.rs
+++ b/crates/harness-agents/src/claude_adapter.rs
@@ -201,10 +201,10 @@ impl AgentAdapter for ClaudeAdapter {
                 continue;
             }
 
-            let event = match parse_stream_json_line(&line) {
-                Some(ev) => ev,
-                None => continue,
-            };
+            let events = parse_stream_json_events(&line);
+            if events.is_empty() {
+                continue;
+            }
 
             if let Some(usage) = parse_stream_json_usage(&line) {
                 if tx.send(AgentEvent::TokenUsage { usage }).await.is_err() {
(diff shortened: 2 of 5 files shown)


majiayu000:
Addressed the two chatgpt-codex-connector findings in fd7b77ca:

- **Single-channel failure reporting**: the stream path no longer emits a `StreamItem::Error` for a terminal result failure — it is reported only via the `Err` return, so the turn lifecycle persists one error instead of two.
- **No contradictory TurnCompleted**: the adapter path tracks terminal result failures and skips the trailing `TurnCompleted` after forwarding the `Error` event.

On the gemini-code-assist suggestion (single-pass tagged-enum deserialization): agreed in principle, but the multi-parse pattern predates this PR and the parser consolidation is tracked in #1792 (unify the two agent traits / one shared parse policy). Keeping this PR scoped to the correctness fix.

PR #1809, opened 2026-07-26
fix(GH1788): strip nested-session markers from the effective child env

### LINKED ISSUE #1788: fix(agents): strip_claude_env misses injected and container-forwarded CLAUDE* variables
## Problem

`strip_claude_env` enumerates the **parent process** env and removes `CLAUDE`-prefixed keys (`crates/harness-agents/src/lib.rs:24`, verified). Three gaps:

1. A `CLAUDE*` var injected via `AgentRequest.env_vars` (applied to the child at `claude.rs:219`, before the strip at `:220`) survives unless the same key also happens to exist in the parent env — so the known SIGTRAP nested-session failure can still reproduce through this path.
2. On the container isolation tier, `CLAUDE*` vars are forwarded as `--env` args (`spawn_contract.rs:89-92`), bypassing the strip entirely.
3. The prefix match is too blunt: it also removes legitimate configuration such as `CLAUDE_CONFIG_DIR` / `CLAUDE_CODE_OAUTH_TOKEN` when present.

## Proposed direction

Strip an explicit allowlist of nesting-detection keys (`CLAUDECODE`, `CLAUDE_CODE_ENTRYPOINT`, ...) from the **effective child environment** — i.e. inside `spawn_contract`, applied to `process_env` and the container `--env` list — rather than from the parent's var list by prefix.

## Acceptance criteria

- A `CLAUDECODE` value injected via request env_vars does not reach the child on host or container paths (regression test).
- Legitimate `CLAUDE_*` config vars pass through.
- Single implementation shared by all spawn paths.

majiayu000:
Closes #1788.

`strip_claude_env` scanned the **parent** process env for `CLAUDE*` keys, so markers injected via `AgentRequest.env_vars` survived (unless the parent coincidentally had the key), and the container tier forwarded `CLAUDE*` vars as `--env` args with no strip at all — both reintroduce the SIGTRAP nested-session failure. The prefix match also deleted legitimate config like `CLAUDE_CONFIG_DIR`.

- explicit allowlist (`CLAUDECODE`, `CLAUDE_CODE_ENTRYPOINT`) now stripped from the effective child env in `spawn_contract` (host + container), plus unconditional `env_remove` in `apply_process_env` for inherited markers
- `execute_review` (which bypasses `prepare_agent_spawn`) strips after `cmd.envs()` instead of before
- legitimate `CLAUDE_*` configuration passes through
- old `strip_claude_env` deleted; 8 call sites collapsed into the contract

Verification: `cargo test -p harness-agents` (203 passed), `cargo clippy --workspace --all-targets -- -D warnings` clean.

### DIFF
--- a/crates/harness-agents/src/claude.rs
+++ b/crates/harness-agents/src/claude.rs
@@ -217,7 +217,6 @@ impl CodeAgent for ClaudeCodeAgent {
         #[cfg(unix)]
         crate::set_process_group(&mut cmd);
         crate::spawn_contract::apply_process_env(&mut cmd, &prepared_spawn);
-        crate::strip_claude_env(&mut cmd);
 
         let _provider_permit = self
             .provider_gate
@@ -376,7 +375,6 @@ impl CodeAgent for ClaudeCodeAgent {
         #[cfg(unix)]
         crate::set_process_group(&mut cmd);
         crate::spawn_contract::apply_process_env(&mut cmd, &prepared_spawn);
-        crate::strip_claude_env(&mut cmd);
 
         // ETXTBSY (error 26) occurs on Linux when a security scanner or indexer
         // briefly opens the executable for writing after it is written. Retry once.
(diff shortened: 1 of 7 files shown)


majiayu000:
Addressed the P1 in bf24a862: the strip list now covers all five wrapper markers the sanitized launcher classifies (`CLAUDECODE`, `CLAUDE_CODE`, `CLAUDE_CODE_ENTRYPOINT`, `CLAUDE_CODE_SESSION_ID`, `CLAUDE_SESSION_ID`), with a cross-reference comment to keep the two lists in sync. Test extended to iterate the full allowlist. Branch also rebased onto main to resolve the test-file conflict with #1807.

PR #1822, opened 2026-07-26
fix(GH1789): bound non-streaming agent output and enforce an idle timeout

### LINKED ISSUE #1789: fix(agents): non-streaming execute() has no timeout and unbounded output buffering
## Problem

The non-streaming execute paths call `ManagedChild::wait_with_output()` with no timeout and accumulate stdout/stderr unbounded in memory (`read_to_end` into `Vec<u8>`, `crates/harness-agents/src/lib.rs:199-228`; call sites `claude.rs:264`, `codex.rs:459`). `stream_timeout_secs` is only consulted on the `execute_stream` paths. Consequences:

- `harness exec` (`harness-cli/src/commands/exec.rs:407`) has **no timeout at all** — a hung agent hangs the CLI forever.
- An agent that dumps a large build log is held entirely in RAM.
- `parallel_dispatch` papers over it with an outer `tokio::time::timeout`, duplicating policy.

## Proposed direction

Enforce the configured timeout inside the non-streaming path itself (same config the streaming path uses); cap or spool captured output beyond a threshold.

## Acceptance criteria

- A never-terminating child fails the non-streaming execute at the configured deadline, and the process group is killed.
- Output capture is bounded (truncation marker or spool-to-file) with a test.
- `harness exec` inherits the timeout without its own wrapper.

majiayu000:
Closes #1789.

Non-streaming `execute()` called `wait_with_output` with no deadline and buffered output unbounded (`read_to_end`) — `harness exec` could hang forever on a stuck agent, and a child dumping a huge log was held entirely in RAM.

- `ManagedChild::wait_with_output` now enforces the agent's `stream_timeout_secs` as an idle deadline on both pipes (same zombie semantics as the streaming path), killing the process group when it fires
- per-stream capture is capped (32 MiB, tail-retained so the trailing result line survives; truncation logged)
- the read loop selects on root exit, so descendants holding pipes are killed immediately rather than stalling reads
- `execute_review` now runs under `ManagedChild` supervision (was a raw `Child`)

Verification: `cargo test -p harness-agents` (218 passed), `cargo clippy --workspace --all-targets -- -D warnings` clean.

### DIFF
--- a/crates/harness-agents/src/claude.rs
+++ b/crates/harness-agents/src/claude.rs
@@ -260,7 +260,8 @@ impl CodeAgent for ClaudeCodeAgent {
         }
         let mut child = crate::ManagedChild::new(child, "claude execute");
 
-        let output = child.wait_with_output().await.map_err(|e| {
+        let limits = crate::OutputLimits::from_stream_timeout_secs(self.stream_timeout_secs);
+        let output = child.wait_with_output(&limits).await.map_err(|e| {
             harness_core::error::HarnessError::AgentExecution(format!(
                 "failed to wait for claude: {e}"
             ))
--- a/crates/harness-agents/src/claude_tests.rs
+++ b/crates/harness-agents/src/claude_tests.rs
@@ -821,6 +821,35 @@ printf '{"type":"result","result":"CLAUDECODE=%s ENTRYPOINT=%s CONFIG_DIR=%s"}\n
 }
 
 #[tokio::test]
+async fn execute_times_out_on_hung_agent() {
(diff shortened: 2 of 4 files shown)


majiayu000:
The failing `event_replay::tests::replay_caller_counts_and_logs_only_applied_writes` check is unrelated to this PR — it is a flaky log-capture test that also fails on main (run 30213212679). Root cause (tracing interest-cache rebuild racing scoped-subscriber registration) diagnosed and fixed in #1828; will rerun CI here once that merges.

PR #1829, opened 2026-07-26
fix(test): bind captured tracing subscribers to the future, not the thread

majiayu000:
## What / Why

Six tests capture tracing output with `tracing::subscriber::set_default(...)` and then `await` across the guard. That guard is **thread-local**. Tokio may migrate a future to a different worker thread at any await point, and when it does, the events the test asserts on are emitted on a thread where the captured subscriber was never installed — so the assertion sees a partial log.

This is not hypothetical. It failed CI on an unrelated PR (#1814, which touches only `harness-agents`):

```
---- event_replay::tests::replay_caller_counts_and_logs_only_applied_writes stdout ----
replay aggregate wording must report exactly the one applied task:
message=event_replay: replayed state superseded by authoritative durable state task_id="replay-superseded"
message=event_replay: compacted log, removed 2 terminal-task line(s)
```

The asserted line `event_replay: applied replayed state to 1 task(s)` is absent while *other* lines from the same call are present — the signature of a subscriber that only saw part of the run. It is scheduling-dependent, not a product defect.

## Change

`WithSubscriber::with_subscriber` attaches the subscriber to the future itself, so it travels with the work across threads:

```rust
// before — guard is thread-local, lost on migration
let updated = {
    let _subscriber_guard = tracing::subscriber::set_default(captured.clone());
    replay_and_recover(&db, &log_path).await?
};

// after — subscriber rides along with the future
let updated = replay_and_recover(&db, &log_path)
    .with_subscriber(captured.clone())
    .await?;
```

Applied to all six sites: `event_replay_tests.rs` (2), `task_db/queries_recovery_tests.rs` (2), `task_runner/store/startup.rs` (2). The last four had the same latent defect and had simply not lost the race yet.

**No assertion is changed** — this repairs how output is captured, not what is required of it (W-12).

## Proof

Characterization, run in this session against a real PostgreSQL 17 instance:

- Before the fix, in isolation: `replay_caller_counts_and_logs_only_applied_writes` passed **5/5**. It only fails under the full parallel suite, which is why it reads as flaky.
- Before the fix, full suite on clean `origin/main`: **1337 passed, 1 failed** — this test.
- CI on #1814 hit the same failure, and the job passed on a plain re-run with no code change.
- After the fix, full suite: **1338 passed, 0 failed**. Additional repeat runs were still in flight when this PR was opened; I will post their results as a comment.
- `cargo clippy -p harness-server --all-targets -- -D warnings` — clean, exit 0.

Note on the evidence: a flaky test cannot be proven fixed by passing runs alone. The argument here rests on the root cause — a thread-local guard held across an await — and on the fix removing that mechanism entirely rather than making the race less likely.

## AI Role

AI-authored. Risk: low — test-only, no production code touched, no assertions relaxed.

## Review Focus

`startup.rs` previously wrapped `TaskStore::open(&db_path)` in an `async` block; it is now the bare future with `.with_subscriber(...)`. Confirm nothing depended on that extra block's laziness — the future is still not polled until the surrounding `select`/`join` drives it.

### DIFF
--- a/crates/harness-server/src/event_replay_tests.rs
+++ b/crates/harness-server/src/event_replay_tests.rs
@@ -3,6 +3,7 @@ use harness_core::types::TaskId;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 use tracing::field::{Field, Visit};
+use tracing::instrument::WithSubscriber;
 use tracing::span::{Attributes, Record};
 use tracing::subscriber::Interest;
 use tracing::{Event, Id, Metadata, Subscriber};
@@ -648,10 +649,9 @@ async fn replay_caller_counts_and_logs_only_applied_writes() -> anyhow::Result<(
     drop(log);
 
     let captured = ReplayCapturedSubscriber::default();
-    let updated = {
-        let _subscriber_guard = tracing::subscriber::set_default(captured.clone());
-        replay_and_recover(&db, &log_path).await?
-    };
+    let updated = replay_and_recover(&db, &log_path)
+        .with_subscriber(captured.clone())
+        .await?;
     assert_eq!(updated, 1, "only the durable replay write is counted");
     assert_eq!(
         db.get("replay-applied")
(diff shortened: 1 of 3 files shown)


majiayu000:
Superseded by #1828, which landed the same fix with the same root cause (callsite interest cached as `never` by a callsite first hit without a capture installed, so later scoped subscribers never see that event).

Recording what this branch established, since it may be useful if the area regresses:

- The first hypothesis on this branch — that `set_default`'s thread-local guard was lost when tokio migrated the future across an await — was **tested and refuted**. Replacing the guard with `WithSubscriber::with_subscriber` left the failure rate unchanged (3/20 vs 1/20 baseline on a targeted co-run).
- The tell that ruled it out: the capture contained *two* events from the same `replay_and_recover` call and was missing a third. A capture that lost its thread would have missed all three.
- Fast local reproduction, about 1 second per attempt: `cargo test -p harness-server --lib -- event_replay queries_recovery` with `HARNESS_DATABASE_URL` set. `queries_recovery_tests` also drives `replay_and_recover`, which is what races the interest cache. The test never fails in isolation.

Closing in favour of #1828.

Issue #1784, opened 2026-07-26
fix(workflow): state-write paths bypass DecisionValidator; shape-fingerprint validation escape hatch


majiayu000:
## Problem

The canonical runtime layer has a well-built `DecisionValidator` (allowlist, terminal-reopen denial, lease ownership, required commands/evidence), but it is not on every write path:

- Of the `upsert_instance_tx` call sites, only the runtime-completion and operator-recovery paths validate first.
- `apply_decision_transition` (`store/instances.rs`) writes a caller-supplied `final_instance` verbatim, checking only `current.state == expected_state` and non-terminality — the transition allowlist and required commands/evidence are never consulted. `store/submission_commit.rs` and `store/coverage_recovery.rs` behave the same.
- `upsert_instance` (`store/instances.rs:25`) is a fully public, unguarded, non-transactional blind overwrite of any instance.
- `is_definition_pin_safety_decision` (`store/runtime_completion.rs:314`, `:348` — verified) accepts a decision **that failed validation** if it structurally matches a fingerprint (`next_state == "blocked"`, two specific commands). Any reducer emitting that exact shape gets an unconditional write past the validator: a policy encoded as a fingerprint rather than a capability.

Note: closed #1715 / #1716 fixed the *legacy* issue-lifecycle and task-db layers and explicitly scoped out this canonical runtime layer.

## Proposed direction

Funnel every state write through a single `commit_transition(tx, instance, decision, context)` that always validates; make `upsert_instance_tx` private; replace the fingerprint escape with an explicit, named `ValidationContext` override; convert the three bypassing paths to call the funnel.

## Acceptance criteria

- No public unvalidated instance write remains; `upsert_instance_tx` unreachable outside the funnel.
- `is_definition_pin_safety_decision` deleted in favour of an explicit context flag.
- Tests: each formerly-bypassing path rejects an allowlist-violating transition.

Evidence from static architecture review of origin/main @ 81c78255; line numbers may drift.

PR #1856, opened 2026-07-29
refactor(workflow): finish declarative reducer migration

### LINKED ISSUE #1793: refactor(workflow): finish the declarative reducer migration and delete the hardcoded reducers
## Problem

The reducer path tries the declarative TOML interpreter first and falls back to ~2,000 LOC of hardcoded per-definition reducers for the same builtin definitions (`reducer.rs:88-100`; `reducer/github_issue_completion.rs` 513, `reducer/pr_feedback_completion.rs` 610, `reducer/prompt_task_completion.rs` 570, plus quality_gate and plan_issue). A 760-LOC test (`state_registry_equivalence_tests.rs`) exists solely to assert the two representations agree. This is a migration that is ~80% done and has stalled with both halves live — every definition change is a two-place edit plus an equivalence-test update, and any divergence is a live behavioral bug.

## Proposed direction

Make the builtin definitions declarative-only; delete the hardcoded reducer chain and the equivalence test; port their behavioral edge cases into fixtures against the interpreter.

## Acceptance criteria

- `reduce_runtime_job_completed` has one path.
- `reducer/*_completion.rs` and the equivalence test are deleted (single biggest LOC reduction available in the crate).
- Interpreter fixture suite covers the edge cases the hardcoded reducers encoded.

majiayu000:
## Summary
- route runtime job completion through the declarative reducer path for built-in workflows
- move built-in completion behavior behind declarative fixtures and delete the hardcoded reducer dispatch chain
- delete the state registry equivalence test now that built-ins are registered declaratively

## Validation
- cargo fmt --all -- --check
- cargo check
- cargo check --workspace --all-targets
- cargo test -p harness-workflow
- cargo test -p harness-server workflow_runtime_worker::prompt_packet -- --nocapture
- cargo test -p harness-server runtime_projection::tests -- --nocapture
- pre-push DB-less checks via env -u HARNESS_DATABASE_URL git push

Closes #1793


majiayu000:
## Repair plan — CI red + branch conflicting

**Status:** `CI Result` FAILURE and `mergeable=CONFLICTING` against current main (branch is from 07-29; the GH-1864 provenance work has since touched the same runtime surface).

**Root cause of the test failure** (run 30466850088, only failure out of 1381):

`workflow_runtime_submission::declarative_tests::declarative_submission_rejects_unknown_and_builtin_ids` panics at `declarative_tests.rs:466`:

```
unexpected error for prompt_task: project '/tmp/…' does not declare workflow definition 'prompt_task'
```

The test asserts that submitting a *built-in* definition id (`prompt_task`) through the declarative path yields the explicit "built-in ids are not allowed here" rejection. After this PR deletes the hardcoded reducers, resolution falls through to the project-declaration lookup first, so the built-in id now surfaces as a generic "not declared by project" error. The guard ordering regressed, not the test: the explicit built-in rejection is the correct user-facing contract (it distinguishes "you used a reserved id" from "typo/missing declaration") and the test protects it per W-12 — fix the code, don't relax the assertion.

**Plan**

1. Rebase onto latest main and resolve conflicts. Note the conflict surface overlaps GH-1864 (PR #1868, declarative pin binding in `state_registry`/`decision_transitions`) — rebase *after* #1868 merges to avoid resolving the same hunks twice.
2. Restore the built-in-id rejection ahead of the project-declaration lookup in the declarative submission path, so reserved ids keep their explicit error while the reducer deletion stands.
3. Re-run: `cargo test -p harness-server --lib` (repro command from CI), then full suite + `cargo clippy --workspace --all-targets -- -D warnings` + `cargo fmt --all`.
4. Re-request review once CI Result is green.

majiayu000:
Rebased onto main (`8908c44d`..) after the GH-1864/GH-1865 series landed.

**Conflict resolution**: one conflict — `state_registry_equivalence_tests.rs` (modify/delete). Main's only change to it was GH-1784's `awaiting_feedback → local_review_gate` expectation line; this PR deletes the harness, so the deletion stands.

**Verification that the deletion is safe here**: the transition allowlist built-ins carry is still `TransitionAllowlist::github_issue_pr_defaults()` (unchanged on this branch, GH-1784's rule included), so no validated capability is withdrawn by the declarative migration. Added `github_issue_pr_builtin_preserves_the_transition_contract` in `state_registry/builtins.rs` — a focused test pinning the full github_issue_pr transition rule set the deleted harness covered, so the contract stays load-bearing without the 760-line equivalence file. The other built-ins had zero main-side churn during the staleness window (only GH-1784 and GH-1864 touched validator.rs; GH-1864 added no allow rules).

Fresh validation on the rebased head: `cargo test -p harness-workflow --lib` 638 passed, `harness-server` runtime_projection (18) + prompt_packet (52) green, workspace clippy `-D warnings` clean.

Issue #1864, opened 2026-08-02
fix(workflow): bind transition authority to immutable workflow provenance


majiayu000:
## Problem

Follow-up to #1784. The validated transition funnel still leaves several provenance boundaries unbound:

- Public canonical initial insertion can create `workflow_instances` without an initial workflow event or decision.
- `apply_decision_transition_with_validator` accepts a caller-supplied `DecisionValidator` that is not bound to the instance's definition id, version, or persisted definition hash.
- Protected instance-field checks omit `data.definition_hash`, although declarative resolution treats it as pinned identity.
- PR-binding repair trusts the latest `BindPr` command regardless of accepted-decision linkage or command status, then performs a same-state write without a repair event/decision.

Evidence on main `65003efa`:

- `crates/harness-workflow/src/runtime/store/instances.rs:13-29`
- `crates/harness-workflow/src/runtime/store/decision_transitions.rs:19-48,69-88`
- `crates/harness-workflow/src/runtime/validator.rs:421-477`
- `crates/harness-workflow/src/runtime/store/pr_binding_repair.rs:34-78`
- `crates/harness-workflow/src/runtime/store/command_facade.rs:4-26`

## Acceptance criteria

- Canonical production instance creation atomically records an initial event; raw insert APIs are private, test-only, or explicitly provenance-bearing.
- A supplied validator carries and verifies exact definition id, definition version, and definition hash before validation.
- `definition_hash` cannot change through decision, same-state, child-start, recovery, or public write paths.
- PR-binding repair accepts only a `BindPr` linked to an accepted decision for the same workflow and records durable repair provenance.
- Regression tests reject cross-definition validators, hash substitution, eventless production creation, and standalone/rejected `BindPr` repair.

majiayu000:
## Fix plan (verified against origin/main `1a6f475b`, 2026-08-04)

Re-verified all evidence before planning. Several cited locations have moved or were renamed; the underlying defects hold. Corrections first.

### Corrections to the issue text

[...]

Issue #1865, opened 2026-08-02
fix(workflow): make decision, command, and prompt provenance immutable across retries


majiayu000:
## Problem

Workflow audit records remain mutable across retries:

- Decision rows overwrite `accepted`, serialized decision data, and rejection reason on id conflict.
- Command dedupe can rewrite payload and decision linkage in place, so one command id can represent different decision generations.
- Content-addressed prompt refs overwrite prompt bytes on conflict instead of proving equality.
- Cancellation cleanup accepts any historical `MarkCancelled` command rather than the marker belonging to the current cancellation decision/generation.

Evidence on main `65003efa`:

- `crates/harness-workflow/src/runtime/store/transaction_helpers.rs:32-54`
- `crates/harness-workflow/src/runtime/store/decisions.rs:3-23`
- `crates/harness-workflow/src/runtime/store/commands.rs:156-226`
- `crates/harness-workflow/src/runtime/store/submission_commit.rs:337-355`
- `crates/harness-workflow/src/runtime/store/prompt_payloads.rs:3-23`
- `crates/harness-workflow/src/runtime/store/runtime_job_state.rs:287-320`

## Acceptance criteria

- Decision id conflicts are exact idempotent replays or errors; accepted state and decision content are append-only.
- Immutable command intent and decision linkage are separated from mutable dispatch status.
- Retried/reactivated commands receive an explicit generation/attempt identity without rewriting historical command provenance.
- Prompt refs are insert-once: equal bytes are idempotent and different bytes return an integrity error.
- Cancellation cleanup is bound to the exact accepted cancellation decision/generation.
- Concurrency and replay tests prove old decisions, commands, prompts, and cancellation markers cannot authorize or overwrite a newer generation.

majiayu000:
## Fix plan (verified against origin/main `1a6f475b`, 2026-08-04)

Re-verified all evidence before planning. Two claims need narrowing; the rest confirmed as written.

### Corrections to the issue text

[...]

PR #1868, opened 2026-08-04
fix(workflow): bind transition authority to immutable provenance (GH-1864)

majiayu000:
Closes two of the four provenance boundaries in #1864. Follow-up to #1784.

## W1 — no eventless creation

`insert_instance_if_absent` and the creation branch of the public `upsert_instance` inserted a workflow row with no record of how it came to exist. Both now record a `WorkflowInstanceCreated` event in the same transaction.

Every creation path is now provenance-bearing:

| Path | Provenance |
|---|---|
| Decision-driven (`create_if_missing`) | its own event + decision record |
| Child start | `ChildWorkflowStarted` |
| `insert_instance_if_absent` / `upsert_instance` | `WorkflowInstanceCreated` (new) |

A losing concurrent insert appends nothing.

## W2 — `data.definition_hash` is protected

The declarative pin decides which definition — and therefore which validator and which legal transitions — govern the instance. It was absent from every protected-field check because it lives inside `data` rather than in a column, so a transition could re-point a workflow at a definition it was never validated against.

It is now compared in `ensure_instance_identity_fields_match`, the chokepoint every row write that compares against a loaded row passes through, so decision commits, same-state writes, recovery, and child start are all covered at once. It is also named explicitly in the decision-transition and public-upsert boundary checks so the rejection reports which field moved.

Three cases are all treated as identity changes:
- substituting the pin
- removing the pin
- introducing a pin on a workflow that never had one

## Not in this PR

The remaining two items from #1864 — binding a supplied `DecisionValidator` to exact definition id/version/hash, and requiring PR-binding repair to link to an accepted decision and record durable repair provenance — are still open.

## Verification

- `harness-workflow` lib: 643 passed
- `harness-server` lib: 1439 passed
- full workspace `cargo test`: exit 0
- `cargo clippy --workspace --all-targets -- -D warnings`: clean
- `cargo fmt --all -- --check`: clean

New tests: `public_creation_records_an_initial_event_atomically`, `apply_decision_transition_rejects_definition_hash_substitution`.

### DIFF
--- a/crates/harness-server/src/workflow_runtime_submission/cancel.rs
+++ b/crates/harness-server/src/workflow_runtime_submission/cancel.rs
@@ -238,7 +238,15 @@ async fn resolve_declarative_cancellation(
         let target_state = cancelled_state(definition.policy(), instance)?;
         return Ok(Some(DeclarativeCancellation {
             target_state,
-            validator: DecisionValidator::new(definition.registered().allowlist.clone()),
+            // The pin resolved to this exact definition, so the validator
+            // carries that identity and the store can re-verify at commit that
+            // it still governs the row it locked (GH-1864).
+            validator: DecisionValidator::for_declarative_definition(
+                &instance.definition_id,
+                definition.definition_version(),
+                definition.definition_hash(),
+                definition.registered().allowlist.clone(),
+            ),
             missing_pin: false,
         }));
     }
@@ -305,7 +313,15 @@ async fn resolve_declarative_cancellation(
     }
     Ok(Some(DeclarativeCancellation {
         target_state: cancelled_state(&policy, instance)?,
-        validator: DecisionValidator::new(definition.registered().allowlist.clone()),
+        // Rebuilt from the persisted snapshot, but only after its version and
+        // content hash were checked against the instance pin above, so the
+        // validator may claim that identity (GH-1864).
+        validator: DecisionValidator::for_declarative_definition(
+            &instance.definition_id,
+            definition.definition_version(),
+            definition.definition_hash(),
+            definition.registered().allowlist.clone(),
+        ),
         missing_pin: true,
     }))
 }
--- a/crates/harness-workflow/src/runtime/mod.rs
+++ b/crates/harness-workflow/src/runtime/mod.rs
@@ -44,6 +44,7 @@ pub mod terminal_state;
 pub mod tier_resolution;
 pub mod transcript;
 pub mod validator;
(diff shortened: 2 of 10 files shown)


majiayu000:
## Review response plan (all three bot findings verified against the diff)

**1. Codex P2 — built-in workflows must be exempt from the hash pin guard: accepted, and it also applies to the follow-up in flight.**
Built-ins are classified non-declarative (`state_registry/versioning.rs:209-212`) and `runtime/tests/declarative_pinning.rs:219-229` documents a built-in instance carrying `definition_hash` as ordinary business metadata. Any unconditional `validator hash == data.definition_hash` comparison therefore rejects valid built-in writes. Fix shape:
- The validator carries an explicit binding `{definition_id, definition_version: Option<u32>, definition_hash: Option<String>}`; only validators resolved from a *declarative* pin populate `definition_version`/`definition_hash`.
- The guard enforces the hash comparison **only when the validator carries a pin**: declarative instance → pinned hash must match exactly (substitution *and* removal both reject); built-in validator (`definition_hash: None`) → `data.definition_hash` is treated as opaque metadata and never compared.
- Regression test: built-in instance with `data.definition_hash = "unrelated-business-metadata"` accepts a same-state write that updates/removes that key.

**2. Gemini medium (`instances.rs:721`) — record `definition_hash` in the `WorkflowInstanceCreated` payload: accepted.**
The pin is the identity anchor this PR introduces, so the creation event must capture it or the audit chain starts one write too late. Will add `definition_hash` (nullable for built-ins) to the payload and assert it in the existing creation-event test.

**3. Gemini medium (`decision_transitions_tests.rs:371`) — cover pin *introduction*: accepted.**
Existing tests cover substitution and removal; will add `apply_decision_transition_rejects_definition_hash_introduction` so all three identity-change directions (introduce / substitute / remove) are proven rejected for declarative instances.

**Sequencing**
1. Push the follow-up commit: `DecisionValidatorBinding` + `ensure_governs` re-verified under the row lock in `apply_decision_transition` (closes the resolve-to-commit window), with the built-in exemption from (1).
2. Add the creation-event payload field (2) and the introduction test (3).
3. Validation: `cargo test -p harness-workflow` + Postgres suites via isolated `HARNESS_DATABASE_URL`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all`.
4. Merge after CI Result is green (squash). #1865's W-items depend on this PR's fenced-write surface, so this lands first.

PR #1873, opened 2026-08-06
fix(workflow): bind cancellation cleanup to the current accepted decision (GH-1865)

majiayu000:
## Summary

GH-1865 W4: cancellation cleanup is bound to the exact accepted cancellation decision/generation.

- `finish_cancellation_cleanup_if_current` accepted **any** historical `MarkCancelled` row. After W2 made attempts supersedeable, that check also honored superseded markers; it likewise honored detached markers (`decision_id` NULL), markers minted for **rejected** decisions, and markers left over from an older generation after a reopen — each could authorize cancelling the current generation's live commands.
- Cleanup now requires the marker to belong to the exact accepted decision that placed the instance in its current state: the latest accepted decision must target the instance's current state and carry a `MarkCancelled` command, and a live (non-superseded) command row must link that decision id to the marker's dedupe key. Anything else returns `NoCancellationCommand` and rolls back.
- The `harness-server` terminal-cancellation retry fixture now commits a real accepted cancellation decision instead of a detached marker, matching the production commit path.

## Test plan

- [x] `cargo test -p harness-workflow --lib` — 653 passed (5 new regression tests: bound happy path, detached, rejected, older-generation, superseded)
- [x] `cargo test -p harness-server --lib terminal_cancellation_retry` — pass
- [x] Full `harness-server` lib suite against `HARNESS_DATABASE_URL` — 1439 passed on the W2 base; re-running on this delta (result before merge)
- [ ] CI

## Notes

- Lands the final acceptance criterion of GH-1865 (W1/W3 in #1871, W2 in #1872). With this, old decisions, commands, prompts, and cancellation markers cannot authorize or overwrite a newer generation.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/harness-server/src/intake/github_coverage_recovery_tests/atomic.rs
+++ b/crates/harness-server/src/intake/github_coverage_recovery_tests/atomic.rs
@@ -71,19 +71,38 @@ async fn cancelled_recovered_quality_gate_is_reactivated_after_terminal_job() ->
         .iter()
         .filter(|record| record.command.requires_runtime_job())
         .collect::<Vec<_>>();
-    assert_eq!(runtime_commands.len(), 1);
-    assert_eq!(runtime_commands[0].id, command.id);
-    assert_eq!(runtime_commands[0].status, WorkflowCommandStatus::Pending);
+    // Recovery re-enqueues the cancelled command as a new attempt rather than
+    // resurrecting the cancelled row, so the cancellation stays in the history
+    // and the retry carries its own identity (GH-1865).
+    assert_eq!(runtime_commands.len(), 2);
+    let superseded = runtime_commands
+        .iter()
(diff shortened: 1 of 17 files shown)


majiayu000:
Review note (2026-08-07, post-merge review): the cleanup binding logic is sound — one minor observation for the record.

`cancellation_marker_is_current_tx` selects the latest accepted decision with `ORDER BY created_at DESC, id DESC`. If two accepted decisions ever share a `created_at` timestamp, the tiebreak falls to `id`, and with random UUID ids that ordering is nondeterministic. Collision probability at Postgres timestamp resolution is very low, so this is a note, not a defect — but if a later change introduces a monotonically increasing decision id (e.g. UUIDv7) or a sequence column, this query becomes fully deterministic for free.

PR #1881, opened 2026-08-06
fix(core): default retention and watchdog sweepers on (GH-1879)

### LINKED ISSUE #1879: Runtime reliability (c): retention and watchdog default off — ship safe defaults on
Split from #1772. **Quick win: config + small code, do this one first.**

## Problem

`runtime_retention_enabled: false`, `task_retention_enabled: false`, `workflow_watchdog_enabled: false` by default — events, decisions, and full transcripts grow without bound, and stuck `blocked`/`awaiting_feedback` workflows are never surfaced. (The orphan-schema reaper's own doc comment records this declaration-execution-gap class already biting once.)

## Proposed direction

Flip the safe defaults: enable retention (with the existing conservative day/batch limits) and the workflow watchdog out of the box, keeping the config knobs for operators to tune or disable. Validate upgrade behavior for existing deployments with large tables (first sweep must be bounded by the batch limits).

majiayu000:
Closes #1879. Part of #1772.

## Summary

- `workflow_watchdog_enabled`, `runtime_retention_enabled`, `task_retention_enabled` now default to **true** — both the `Default` impl and the serde `#[serde(default)]` path, so existing configs that never mention the flags pick up the sweepers on upgrade.
- Bounds are unchanged and conservative: 30-day age minimum, 1000 rows per hourly tick, only terminal families/tasks pruned (active, pending, dependency-blocked, resumable are never touched).
- Watchdog stays read-only (reports/logs stuck workflows only).
- Ops doc updated: upgrade opt-out is `*_enabled: false` in `WORKFLOW.md` storage config.

## Behavior-change note

Upgrading deployments with >30-day terminal history will see gradual pruning (bounded per tick). Operators who want to keep history must set the flags to `false` **before** upgrading.

## Validation

- `cargo test -p harness-core --lib` — 800 passed

### DIFF
--- a/crates/harness-core/src/config/workflow/storage.rs
+++ b/crates/harness-core/src/config/workflow/storage.rs
@@ -19,9 +19,9 @@ pub struct WorkflowStoragePolicy {
     /// Maximum legacy unregistered path-derived schemas to drop per tick.
     #[serde(default = "default_orphan_reaper_legacy_batch")]
     pub orphan_reaper_legacy_batch: usize,
-    /// Enable workflow-runtime stuck-instance reporting. Off by default so
-    /// existing deployments keep their current background behavior on upgrade.
-    #[serde(default)]
+    /// Enable workflow-runtime stuck-instance reporting. Read-only (reports
+    /// and logs only), so it ships on; disable to silence the signal.
+    #[serde(default = "default_true")]
     pub workflow_watchdog_enabled: bool,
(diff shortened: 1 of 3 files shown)


majiayu000:
Closing in favor of #1884 — same fix for GH-1879 with green CI (this PR has failing Test / CI Result checks and both target the same config defaults change).

Issue #1877, opened 2026-08-06
Runtime reliability (a): dual lease systems can disagree — bind job lease to workspace lease


majiayu000:
Split from #1772.

## Problem

Runtime jobs lease in `runtime_jobs` (owner + `lease_generation`, TTL 600s); workspaces lease in `workspace_leases` (owner_session + PID liveness). Nothing ties them atomically. A worker whose job lease expires mid-turn keeps its agent process and workspace lease running while a second worker can claim the same job and re-run against a dirty tree.

Related decoy: `WorkflowInstance.version` is serialized and named like a CAS token, but every mutation relies on `SELECT ... FOR UPDATE` + string `expected_state`; any future writer using `version` optimistically will silently lose updates.

## Proposed direction

Bind job lease renewal to workspace lease (single ownership epoch). Either enforce version-CAS or remove/rename the field so its shape stops inviting optimistic writes.

Full analysis: `docs/references/runtime-reliability.md`.

PR #1887, opened 2026-08-07
feat(workflow): dead-letter lease-expired activity completions

### LINKED ISSUE #1878: Runtime reliability (b): lease-expired completions are silently dropped — route to dead-letter/reconcile
Split from #1772.

## Problem

When a worker's job lease expires mid-turn, `commit_runtime_activity_completion_with_transcript_if_owned` returns `None` and the worker `tracing::warn!`s and drops the entire result. Completed agent work — potentially a full successful turn with artifacts — vanishes with no durable record, no retry signal, and no operator visibility.

## Proposed direction

Route lease-expired completions to a dead-letter/reconcile path instead of dropping: persist the completion payload durably, surface it to the reducer/operator, and let reconciliation decide whether the result can still be applied or must be re-run.

majiayu000:
Closes GH-1878

## Summary

When a worker's job lease expired mid-turn, `commit_runtime_activity_completion_with_transcript_if_owned` returned `None` and the worker dropped the completed result — a full successful turn with artifacts could vanish with no durable record, no retry signal, and no operator visibility.

- New `runtime_job_completions_dlq` table (store migration v26): `runtime_job_id`, `owner`, `lease_expires_at`, `result` (JSONB), `transcript` (JSONB), `recorded_at`, `applied`
- `record_lease_expired_completion` store method persists the rejected payload and emits a `LeaseExpiredCompletionRecorded` runtime event (operator-visible)
- Local worker now records to the DLQ instead of silently dropping; the job stays running and reclaimable, so dispatch behavior is unchanged
- Remote-host path already surfaced lease rejection as 409 to the caller — untouched

## Verified

- New DB test: DLQ row durable + unapplied, runtime event emitted, job remains reclaimable
- Full harness-workflow DB suite: 649 pass (local Postgres)

## Notes

Reconciliation that decides apply-vs-rerun is deliberately out of scope — it needs the ownership-epoch work of GH-1877 to be safe.

### DIFF
--- a/crates/harness-workflow/src/runtime/store/activity_completion.rs
+++ b/crates/harness-workflow/src/runtime/store/activity_completion.rs
@@ -1,4 +1,5 @@
 use super::*;
+use uuid::Uuid;
 
 impl WorkflowRuntimeStore {
     pub async fn complete_runtime_job_if_owned(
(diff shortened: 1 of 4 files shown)


majiayu000 (commented):
Self-review of the full diff (213 lines, 4 files).

**Findings (all addressed in `511ea0d0`):**

1. **P2 — DLQ was not idempotent.** The same job re-expiring (worker crash → restart → re-execute → re-expiry) inserted a new DLQ row each time, which would confuse reconciliation. Now keyed on `runtime_job_id` (one pending record per job, first wins via `ON CONFLICT DO NOTHING`), and the duplicate runtime event is skipped.
2. **P3 — transcript persistence path had no test.** Added `lease_expired_completion_persists_transcript_payload`, verifying the transcript record (schema, producer job ref, workflow id) lands in the DLQ.

**Checked and fine:**
- `complete_runtime_job_if_owned` (non-transcript variant) also returns `None` on lease mismatch, but it has no production callers — only the worker's `_with_transcript` path runs, which is the one wired to the DLQ.
- The remote-host 409 path already surfaces lease rejection to the caller; only the local worker dropped silently.
- Migration v26 is additive and idempotent (`IF NOT EXISTS`); no existing-table rewrites.
- DLQ insert and runtime event are not atomic; if the event write fails the row is still durable and the worker logs at error level — acceptable, main asset is the payload.

CI re-running for `511ea0d0`.

PR #1889, opened 2026-08-07
fix(workflow): cancel agent and release workspace when job lease is lost

### LINKED ISSUE #1877: Runtime reliability (a): dual lease systems can disagree — bind job lease to workspace lease

majiayu000:
Closes GH-1877

## Summary

When a worker's job lease expired mid-turn, `execute_with_lease_renewal` **dropped the execution future**: the agent process kept running, `finish_runtime_workspace` never ran (workspace leaked), and a second worker claiming the job re-ran against a dirty tree. The workspace lease could not prevent this — workflow workers share one server process, so `owner_session` is identical for both workers.

**Fix:**
- `RuntimeJobExecutor::cancel_execution` (default no-op): on lease loss the worker signals the executor via a per-executor `Notify` (race-free: executor instances are per-worker-tick)
- The turn loop watches the signal, `adapter.interrupt()` terminates the agent child, execute returns a cancelled result, and the normal workspace finalization releases the workspace **before** a reclaimer can touch it
- The worker waits up to 30s for that cleanup instead of dropping the future
- GH-1878's DLQ remains the backstop for the result payload; the workspace reaper backs up cleanup

**Cluster B (same issue):** `WorkflowInstance.version` is documented as informational — a mutation counter, not a CAS token. Every write goes through `SELECT FOR UPDATE` + `expected_state`; the doc comment prevents future optimistic writes.

## Verified

- New lease-tamper worker test: tamper steals the lease mid-execution → cancel invoked once, execution awaited (not dropped), result lands in DLQ
- harness-workflow DB suite: 651 pass (local Postgres)

## Notes

Live multi-worker contention against a real adapter is not covered by CI; the lease-tamper test exercises the exact code path deterministically.

### DIFF
--- a/crates/harness-server/src/workflow_runtime_worker/executor.rs
+++ b/crates/harness-server/src/workflow_runtime_worker/executor.rs
@@ -48,20 +48,22 @@ use structured_output::{
 };
 pub(super) struct ServerRuntimeJobExecutor<'a> {
     pub(super) state: &'a Arc<AppState>,
-    /// Fired when the owning runtime job lease is lost mid-turn so the turn
-    /// loop interrupts the agent and the workspace cleanup can run (GH-1877).
-    lease_lost: Arc<tokio::sync::Notify>,
+    /// Stateful lease-lost signal: `watch` keeps the latest value, so a
+    /// cancellation that fires before the turn loop starts polling is not
+    /// lost (GH-1877).
+    lease_lost: Arc<tokio::sync::watch::Sender<bool>>,
 }
 impl<'a> ServerRuntimeJobExecutor<'a> {
     pub(super) fn new(state: &'a Arc<AppState>) -> Self {
+        let (lease_lost, _) = tokio::sync::watch::channel(false);
         Self {
             state,
-            lease_lost: Arc::new(tokio::sync::Notify::new()),
+            lease_lost: Arc::new(lease_lost),
         }
     }
 
     pub(super) fn cancel_lease_lost(&self) {
-        self.lease_lost.notify_waiters();
+        let _ = self.lease_lost.send(true);
     }
 
     pub(super) async fn execute_inner(&self, job: RuntimeJob) -> anyhow::Result<ActivityResult> {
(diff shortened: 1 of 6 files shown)


majiayu000 (commented):
Self-review of the full diff (251 lines, 6 files).

**Findings (both addressed in `51718d71`):**

1. **P1 — lease-lost signal could be lost before the turn loop started polling.** `Notify::notify_waiters()` only wakes current waiters: a cancellation fired while the prompt packet was still being built (before the turn loop polled the branch) was dropped, the agent never got interrupted, and the 30s grace timeout then dropped the future — regressing to the original leak. Now a stateful `watch` channel; a pre-fired signal is still observed on first `changed()`.
2. **P2 — the new turn-lifecycle test was flaky (3-in-8 failures).** The adapter's `start_turn` returned immediately, so `select` could complete execution before the lease-lost branch fired (guard `execution_result.is_none()`). The adapter now blocks until `interrupt` fires, making the lease-lost branch deterministically win; verified stable over 5 consecutive runs.

**Checked and fine:**
- Worker waits (30s grace) for executor cleanup instead of dropping the future; on timeout it returns a failed result and the DLQ (GH-1878) plus workspace reaper are the backstops.
- `cancel_execution` has a default no-op so other `RuntimeJobExecutor` impls (tests) are unaffected; executor instances are per-worker-tick, so the single watch channel is race-free by construction.
- The interrupt branch mirrors the existing timeout/stall branches (break `'outer` after setting `execution_result`); the adapter interrupt terminates the agent child and the normal workspace finalization path runs after `run_turn_lifecycle` returns.
- `WorkflowInstance.version` doc comment (Cluster B) matches actual usage — it feeds coverage recovery but every write is `FOR UPDATE` + `expected_state`.

CI re-running for `51718d71`.

PR #1890, opened 2026-08-07
feat(server): supervise background loops and surface config-parse failures

### LINKED ISSUE #1880: Runtime reliability (d): unsupervised background loops and silent config-parse degradation
Split from #1772.

## Problem

~10 bare `tokio::spawn` loops from `http/mod.rs` with warn-and-continue failure handling, no leader election, and `WORKFLOW.md` re-parsed every tick — a malformed config file silently disables retention, watchdog, and reaping simultaneously while dispatch keeps running.

## Proposed direction

Supervised background tasks with health surfacing (structured status per loop, liveness exposed to operators), and config-parse failure treated as loud degraded mode (surfaced error + last-known-good config) rather than silent feature loss.

majiayu000:
Closes GH-1880

## Summary

~10 bare `tokio::spawn` orchestration loops re-parsed WORKFLOW.md every tick with warn-and-continue failure handling — a malformed config **silently** disabled retention, watchdog, and reaping (the orphan reaper's failure arm had no log at all) while dispatch kept spending.

- New `BackgroundLoopHealth` registry: every safety loop registers a handle and reports tick outcomes
- **operator-monitor health now exposes**: per-loop liveness (`background_loops`: tick counts, failures, last-tick/last-success ages, staleness past 15 min) and one aggregated `config_parse_failure` (first/last seen, occurrences, affected loops, last error)
- Stale loops and active config-parse failure flag `degraded_subsystems` → operator health status degrades
- All config-reload failure arms converge on a shared loader: error-level log with loop name, aggregated health signal, cleared on recovery
- The orphan reaper's silent `Err(_)` arm is now loud

## Verified

- `loop_health` unit tests: tick tracking, config-failure aggregation + clear, staleness
- Full harness-server DB suite: **1452 pass** (local Postgres)
- Legacy REST inventory refreshed for the OperatorHealth shape change (guard test)

## Notes

Leader election is explicitly out of scope (multi-instance singleton safety stays advisory-lock territory); per-loop restart supervision was rejected — warn-and-continue loops are self-healing, the registry makes a dead loop visible instead.


majiayu000 (commented):
Self-review of the full diff (702 lines, 16 files).

**Findings (both addressed in `b76c175e`):**

1. **P1 — `Box::leak` per snapshot in the degraded-subsystems fallback.** The unknown-loop arm leaked a static string on every operator-monitor request while the loop stayed stale. Now a constant `"background_loop_stale"`; the precise name is already exposed in `background_loops`.
2. **P2 — `tick_failed` discarded the error and double-counted ticks.** The shared loader already calls `tick_ok` per iteration, so a business failure then counted the tick twice; `tick_count` now counts successful ticks only, failures advance `failure_count` and keep `last_error` so operators see the reason.

**Checked and fine:**
- All four config-reload failure arms converge on the shared loader (error-level log with loop name); the orphan reaper's previously silent `Err(_)` arm is now loud.
- `clear_config_failure` on any successful parse is correct: all loops load the same WORKFLOW.md, so one success proves recovery.
- Staleness threshold (15 min) vs loop intervals (300s/3600s) is conservative — a retention loop that dies at the top of its hour is flagged within 15 minutes.
- `LegacyQueueCounts`-style inventory guard: OperatorHealth shape change refreshed once; internal `LoopSnapshot` changes don't affect the inventory (it tracks DTO hashes).
- Leader election intentionally out of scope; per-loop restart rejected (warn-and-continue loops self-heal, the registry makes death visible).

CI re-running for `b76c175e`.

PR #1891, opened 2026-08-07
fix(server): unify background-loop tick reporting semantics (GH-1880)

majiayu000:
Review fixes for the GH-1880 background-loop health reporting landed in #1890.

**Problems found in review:**
1. load_workflow_config_for_loop recorded tick_ok on config load success — config load is not a completed tick. The reaper counted 2 ticks per iteration while retention counted config loads, so tick_count was incomparable across loops.
2. Config-disabled loops (watchdog/retention/reaper) never reported any tick → permanently stale in operator health, falsely degrading /health 15 minutes after a config that disables them.
3. Loop-body errors were invisible: reaper reap failure, retention/task prune failures were only traced, never surfaced in the registry.
4. operator_monitor called snapshot() and config_failure_snapshot() twice per request.

**Changes:**
- load_workflow_config_for_loop: config success only clears the aggregated config failure; tick reporting is each loop's responsibility
- Each loop reports tick_ok at its work-done point (disabled-by-config counts as a healthy no-op tick)
- Reaper/retention/task-retention/watchdog report tick_failed on loop-body errors
- Watchdog aggregates the first scan error into one tick result
- operator_monitor takes single snapshots

**Verified:** cargo fmt --check, cargo clippy -p harness-server --all-targets -- -D warnings, cargo test -p harness-server --lib operator_monitor (35/35), loop_health tests (3/3).

### DIFF
--- a/crates/harness-server/src/handlers/operator_monitor.rs
+++ b/crates/harness-server/src/handlers/operator_monitor.rs
@@ -221,6 +221,8 @@ async fn build_operator_monitor(state: &AppState) -> anyhow::Result<OperatorMoni
         .workspace_mgr
         .as_deref()
         .map(|manager| manager.live_count());
+    let loop_snapshots = state.background_loops.snapshot(BACKGROUND_LOOP_STALE_SECS);
+    let config_parse_failure = state.background_loops.config_failure_snapshot();
 
     Ok(OperatorMonitorPayload {
         generated_at: generated_at.to_rfc3339(),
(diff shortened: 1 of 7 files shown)


majiayu000 (commented):
Reviewed in place of the Codex review bot (quota-limited). Two-pass review of the full diff (7 files, +75/-44 against main).

**Findings addressed in this PR:**
- P1: config-load success was counted as a completed tick (reaper double-counted, retention counted config loads) — tick reporting now happens at each loop's work-done point
- P1: config-disabled loops never ticked → permanently stale → false /health degradation — disabled-by-config now reports a healthy no-op tick
- P2: loop-body errors (reap/prune/dry-run/scans) were trace-only, invisible in the registry — now tick_failed
- P2: watchdog kept the last of multiple scan errors in one tick — now keeps the first, matching the stated contract
- P3: operator_monitor took two snapshots per request — now one

**Verified:** fmt, clippy -D warnings, operator_monitor 35/35 with a disposable Postgres DB, loop_health 3/3.

No remaining blocking findings.

Issue #1934, opened 2026-08-12
fix(runtime): harden repository execution and lease safety


majiayu000:
## Problem

Repository-level execution can admit stale GitHub subjects, allow work to outlive terminal workflow transitions, and rely on runtime leases that are insufficiently fenced across renewal, reclaim, and completion races.

The repository single-writer contract also needs a PostgreSQL-backed lock that coordinates across Harness processes rather than only within one process.

## Expected behavior

- Verify the exact GitHub issue or pull request is still open before persisting execution work.
- Atomically cancel unfinished commands and runtime jobs when a workflow becomes terminal.
- Fence remote-host renewal and completion with generation, expiry, and a lease-scoped proof.
- Preserve migration compatibility for leases already running before proof issuance.
- Hold repository single-writer ownership through a dedicated PostgreSQL advisory-lock connection.
- Fail closed when durable runtime state or the PostgreSQL lease store is unavailable.
- Keep completion, workflow events, transcripts, and stale-completion dead letters transactionally consistent.

## Proposed implementation

The relevant paths have been traced through the workflow runtime store, runtime-host handlers, execution admission, and workspace pool. The implementation will:

1. add the remote lease-proof migration and compatibility path;
2. make completion and stale-completion persistence atomic;
3. add terminal-state enqueue fencing and in-flight cancellation;
4. add exact GitHub subject-state validation;
5. add a global PostgreSQL repository write lease;
6. cover migration, race, recovery, and API behavior with PostgreSQL-backed tests.

The implementation will remain in a Draft PR until the protocol and migration details are ready for maintainer review.

majiayu000:
I traced the root causes to four execution boundaries: admission, terminal transition, remote lease ownership, and repository workspace ownership.\n\nBefore opening the Draft PR, the implementation direction is:\n\n- fail closed on an exact GitHub issue/PR open-state check before persistence;\n- atomically fence enqueue and cancel unfinished runtime work on terminal transitions;\n- use generation + expiry + lease-scoped proof for remote renewal/completion, with a bounded v27 compatibility path for already-running legacy leases;\n- keep completion events, transcript persistence, and stale-completion dead letters within their owning transactions;\n- hold a PostgreSQL advisory lock on a dedicated connection for single-writer repositories.\n\nThe implementation and PostgreSQL race/migration tests are complete locally. I am proceeding with a Draft PR so the protocol and migration behavior remain reviewable before marking it ready.

majiayu000:
[...]

Issue #1794, opened 2026-07-26
refactor(workflow): replace the process-global definition registry with an injected handle


majiayu000:
## Problem

`static REGISTRY: OnceLock<RwLock<WorkflowDefinitionRegistry>>` (`runtime/state_registry.rs:363`) is accessed through 16 free functions that each `.expect("workflow definition registry lock poisoned")`. Consequences:

- `WorkflowInstance::is_terminal()` routes through the registry, so **every terminality check takes a blocking `std::sync::RwLock` read** — including inside open Postgres transactions on tokio worker threads (`store/instances.rs:155`, `store/commands.rs:84`).
- One panic while the write lock is held poisons the lock; every subsequent workflow operation in the process then panics.
- Registration is global-per-process, so tests can't register isolated definitions in parallel — hence `freeze_workflow_definition_registry()`.

## Proposed direction

Thread `Arc<WorkflowDefinitionRegistry>` through `WorkflowRuntimeStore` (already passed everywhere). Delete the global, the 16 free functions, the poison expects, and the freeze mechanism.

## Acceptance criteria

- No `OnceLock`/global registry; definition lookup goes through the store's handle.
- No blocking lock acquisition inside open DB transactions on the terminality path.
- Tests register isolated definitions concurrently without a freeze step.

Evidence from static architecture review of origin/main @ 81c78255; line numbers may drift.

PR #1966, opened 2026-08-17
refactor(cli): keep commands.rs as thin clap dispatch

### LINKED ISSUE #1954: refactor(cli): unify the split cmd/ vs commands/ module trees and inline dispatch logic
## Problem

`harness-cli` has no single convention for where subcommand implementations live:

- `src/commands/` holds `serve.rs`, `exec.rs`, `status.rs`, `runtime.rs`, `reconcile.rs`, `eval.rs`
- `src/cmd/` holds `mcp_server.rs`, `pr.rs` (a parallel tree)
- `src/gc.rs` sits at the crate root, outside both trees
- Rule/skill/plan/execpolicy handling is inlined in `commands.rs` (~1700 lines) inside the giant `run()` match instead of living in either tree

## Why it matters

New subcommands land wherever is convenient, which is how `gc` escaped and `pr`/`mcp_server` ended up in a parallel tree. The dump-file `commands.rs` keeps growing.

## Suggested direction

Pick one tree (`commands/`), move `cmd/*`, `gc.rs`, and the inline rule/skill/plan/execpolicy logic into it, and keep `commands.rs` as thin clap dispatch only.

Found during an architecture review of origin/main (9634e5a9).

majiayu000:
## What

Keep `commands.rs` as thin clap dispatch after `#1962`.

- Move runtime-log / tracing bootstrap into `commands/logging.rs`.
- Move config discovery, capability-profile detection, and central `WORKFLOW.md` install into `commands/config.rs`.
- Move `Gc` / `Rule` / `ExecPolicy` / `Skill` / `Plan` / `Pr` / `Runtime` clap types into their existing modules.
- Move exec-policy/actor/sandbox unit tests next to `exec.rs`.

`commands.rs` now owns the top-level `Cli` / `Command` parser and the `run()` match only.

Closes #1954

## Why

`#1962` put every subcommand under `commands/`, but `commands.rs` was still a 1532-line clap + logging + dispatch file. This is the remaining `#1954` slice.

## Validation

- `cargo check -p harness-cli --all-targets`
- `cargo test -p harness-cli --bin harness` (99 passed)
- pre-commit hook: fmt + scoped clippy
- pre-push workspace lib tests were blocked by pre-existing hanging `harness-agents` `github_app_issuer_*` mock-HTTP tests (unrelated to this CLI move). CI remains the merge gate.

## Out of scope

- `#1794` (injected workflow registry)
- `#1934` safety slices
- flattening `Serve` / `Exec` / `Status` field variants into `*Args` structs

### DIFF
--- a/crates/harness-cli/src/commands.rs
+++ b/crates/harness-cli/src/commands.rs
@@ -1,22 +1,12 @@
-#[cfg(feature = "server")]
-use chrono::{DateTime, NaiveDateTime, Utc};
-use clap::{ArgAction, Args, Parser, Subcommand};
-#[cfg(feature = "server")]
-use harness_server::server::RuntimeLogMetadata;
-#[cfg(feature = "server")]
-use std::cmp::Ordering;
-#[cfg(feature = "server")]
-use std::fs::OpenOptions;
-use std::fs::{self, File};
-use std::io::{self, Write};
-use std::path::{Path, PathBuf};
-use std::sync::{Arc, Mutex};
-use tracing_subscriber::fmt::writer::MakeWriter;
+use clap::{ArgAction, Parser, Subcommand};
+use std::path::PathBuf;
 
+mod config;
 mod eval;
 mod exec;
 mod execpolicy;
 mod gc;
+mod logging;
 mod mcp_server;
 mod plan;
 mod pr;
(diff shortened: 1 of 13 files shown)


majiayu000:
Fresh-context review of exact HEAD 92ca79093a6380bf96ff68d3945f12bdb317019c completed.

No blocking findings. The clap surface, module visibility, config/logging behavior, moved tests, and feature-gated builds remain compatible. Exact-head CI Result, Clippy, Test, Security Audit, and the no-server harness-cli compile check are green.

The tempfile suggestion is valid cleanup but non-blocking here: the affected tests used the same manual temporary-directory pattern on origin/main and this PR only relocates them. Keeping that cleanup out of this structural-only diff preserves the reviewed behavior boundary.

APPROVED

PR #1985, opened 2026-08-22
fix(server): keep websocket heartbeat live during request dispatch

### LINKED ISSUE #1984: server: WebSocket heartbeat can close healthy connections while a long RPC is being handled
## Problem

In `crates/harness-server/src/websocket.rs` (`handle_socket`), the JSON-RPC request handler is awaited **inside** the `tokio::select!` read arm:

```rust
msg = ws_stream.next() => {
    ...
    Ok(req) => router::handle_request(&state, req).await,
```

While a slow RPC runs (agent-facing calls routinely take seconds to minutes), the select loop cannot service other branches:

1. Heartbeat Ping is sent, `pong_pending = true`.
2. Client answers Pong promptly — but the frame sits unread because the loop is inside `handle_request`.
3. The next heartbeat tick becomes due. When the handler finally returns, both the buffered Pong and the tick are ready; `tokio::select!` picks randomly, and the tick branch sees `pong_pending == true` → **healthy connection closed as stale**, mid-session.

Any request whose duration crosses a heartbeat boundary has up to ~50% chance of killing the connection even when the client is perfectly responsive.

## Proposed fix

Move request dispatch onto a serialized worker task fed by a bounded channel:

- Main select loop only reads frames and forwards request texts via `try_send` — Pong frames and heartbeat ticks are serviced within microseconds regardless of handler duration.
- A single worker preserves today's strict FIFO response ordering (no concurrent dispatch semantics change).
- Bounded backlog (e.g. 16): when full, respond with a JSON-RPC error instead of growing memory without bound or stalling the heartbeat (fail-closed backpressure, consistent with the alerting B-011 pattern).
- Worker is aborted alongside `send_task`/`notif_task` when the connection closes.

## Regression tests

The existing tests drive a real tungstenite client against a bound listener with configurable heartbeat interval. To reproduce deterministically the loop needs an injectable slow responder, so the per-connection loop should take the request handler as a closure (production passes `router::handle_request`, tests pass a sleeping one):

1. Client answers every Ping; a request whose handler sleeps > 2×interval → connection must survive.
2. Silent client → connection closed after missed pong (existing behavior preserved).

majiayu000:
## Problem

JSON-RPC request handlers ran inline inside the connection loop's `tokio::select!` arm (`websocket.rs`). A slow handler — a long agent call, a blocked store query — delayed Pong servicing and heartbeat ticks. When the handler returned, tokio's random selection among ready branches could observe `pong_pending` still set from an earlier Ping and kill a **healthy** connection that had answered every Pong it was given the chance to answer. The longer handlers run, the higher the false-positive risk; with multi-minute agent turns this is close to guaranteed disconnects.

Closes #1984.

## Fix

Move dispatch off the connection loop onto a serialized per-connection worker task:

- `websocket_dispatch::dispatch_requests` handles queued requests one at a time (preserving today's strict FIFO response ordering), but on its own task — the connection loop only forwards frames, services Pongs, and ticks the heartbeat.
- Requests are forwarded through a bounded channel (`WS_REQUEST_BACKLOG = 16`). When the backlog is full the server rejects with a JSON-RPC internal error instead of growing memory without bound.
- The worker is aborted alongside `send_task`/`notif_task` when the loop exits.

Housekeeping: the shared test fixture (`AppState` builder + sandbox-tolerant TCP/WS client helpers) moved out of `websocket.rs` into `websocket_test_support.rs`, keeping both files under the size limit while letting the new dispatch-worker tests reuse the same fixture. One stale env-audit allowlist entry moved with it.

## Verification

Local throwaway PostgreSQL (port 5433), fresh runs:

- `cargo fmt --all --check` ✅
- `cargo clippy --workspace --all-targets -- -D warnings` ✅
- `cargo test -p harness-server --lib`: **1629 passed, 0 failed** ✅
- `cargo test -p harness-core --lib`: 681 passed (env audit updated accordingly) ✅
- New tests: pipelined requests answered in order over the real socket (E2E); dispatch-worker ordering + parse-error resilience (unit).

Known limitation, noted on the issue: a deterministic E2E reproduction of the exact heartbeat race needs an injectable slow responder — no production RPC handler sleeps controllably today. The structural fix removes the stall window outright; handler-injection can be a follow-up if we want belt-and-braces regression coverage.


majiayu000:
Pushed 5885c01f addressing the review findings:

- **Storage Legacy Openers failure**: `make_test_state_with_config` now carries an explicit `#[cfg(test)]` — the guard's scanner doesn't see the module-level `#![cfg(test)]` in lib.rs, but does honor per-item cfg(test) as legitimate test context. Guard passes locally; no allowlist changes.
- **Shutdown abort**: teardown now closes the request queue and gives the dispatch worker a bounded 5s window (`WS_DISPATCH_DRAIN_GRACE`) to finish its in-flight handler and drain queued requests, then gives the sender task a bounded window to flush pending frames before the sink drops. Cancels only on grace expiry.
- **Backlog rejections with id:null**: rejection responses now echo the request id mined from the raw frame via `codec::decode_request` (unparseable frames keep id null). Outbound writes remain non-blocking (unbounded-channel send, drop-on-closed).

majiayu000:
Pushed c446e518: integration merge with main. #2000's module split (mod/connection/origin) and #1996 bounded-outbound landed first, so this resolves the semantic conflict by keeping the serialized dispatch worker (main's inline dispatch would have re-broken GH-1984 — long calls starved heartbeat ticks) and adopting #1996's bounded outbound queue + slow-client close on top: `dispatch_requests` now writes into the bounded queue and raises the backlog-close flag on overflow; teardown keeps the bounded graceful drain. Env-audit allowlist resolved as main's set plus the websocket_test_support.rs swap (helpers consolidated back there).

Verified against a disposable local PostgreSQL 16: websocket suite 15/15 including `websocket_closes_connection_when_outbound_backlog_overflows` (#1996) and `websocket_pipelined_requests_answered_in_order` (GH-1984); raw_process_env_reads audit and storage-opener guard green; workspace clippy -D warnings + fmt clean.

PR #1995, opened 2026-08-22
perf(server): push prompt-family kind filters into submission listing SQL

### LINKED ISSUE #1994: perf: submission listing pages the whole table when filtering by Review/Planner/Prompt kind; row projection computed 3x
## Problem

`append_runtime_definition_summaries` (`crates/harness-server/src/http/task_query_routes/runtime_submissions.rs:47-138`) lists runtime submissions with keyset pagination, but one slice of the kind filter never reaches SQL:

### 1. Fine kind discrimination is post-fetch → full-table paging

The store query already discriminates coarsely (`submission_instances.rs:120-126`):

- `$7` (include_issue): `github_issue_pr` rows **with** `issue_number`
- `$8` (include_pr): `github_issue_pr` rows **without** `issue_number`
- `$9` (include_prompt): everything that is not `github_issue_pr`

But Prompt vs Review vs Planner is decided per-row in Rust via `runtime_submission_task_kind` → `prompt_execution_policy(&workflow.data)` reading `data.execution_policy.task_kind` (`runtime_submissions.rs:108`, `workflow_runtime_submission/mod.rs:122`). When the caller asks for e.g. `kinds=[Planner]` or `kinds=[Review]`, SQL still returns the entire prompt family and the loop discards non-matching rows page by page — on a table dominated by plain Prompt submissions, listing Reviews walks the whole table (fetch 50, project every row, throw most away, repeat) until it accumulates `limit` matches or hits the end.

The discriminator is available in SQL as `data->'data'->'execution_policy'->>'task_kind'` (serde `snake_case`: `prompt`/`review`/`planner`; missing policy defaults to `prompt`), so the filter can be pushed down like `task_statuses` already is (`submission_instances.rs:128`).

### 2. Row projection computed up to 3×

`RuntimeWorkflowProjection::from_workflow_with_registry(...)` runs:

- once per fetched workflow for the cursor candidate (`:91-98`),
- again in the loop body for every workflow (`:100-104`),
- a third time inside `runtime_workflow_task_summary` (`:115-120` → `:163-169`).

So each row is projected twice and the last row of every page three times.

## Proposed fix

1. Extend `WorkflowSubmissionFilter` with `prompt_task_kinds: Vec<String>` (empty = unconstrained). Add one predicate to `list_submission_instances_page`:
   `(NOT $9 OR cardinality($16::text[]) = 0 OR COALESCE(data->'data'->'execution_policy'->>'task_kind', 'prompt') = ANY($16))`.
   The handler maps `filter.kinds ∩ {Prompt, Review, Planner}` into it. Rows with malformed `execution_policy` keep today's fail-closed behavior because `runtime_submission_task_kind` still validates each fetched row before inclusion — SQL narrowing only removes rows that could never match anyway.
2. Compute the projection once per row and reuse it for the kind check, dedupe key, cursor candidate, and summary construction (change `runtime_workflow_task_summary` to accept the precomputed projection / borrow the instance).

No behavior change for callers: same ordering, same dedupe, same error surface.

majiayu000:
Closes #1994

## Problem

Listing submissions filtered to `Review` or `Planner` paged through the **entire prompt family**. The store query only discriminates coarsely (Issue rows / Pr rows / non-github prompt-family rows); the Prompt vs Review vs Planner split was applied per row in Rust after fetch, reading each row's `execution_policy.task_kind`. On a Prompt-dominated table, a Review filter fetched page after page (50 at a time), projected every row, and discarded most of them — O(table) work per listing request.

Separately, every row's `RuntimeWorkflowProjection` was computed up to three times: once for the cursor candidate, once in the loop body, and once more inside the summary builder.

## Change

- `WorkflowSubmissionFilter` gains `prompt_task_kinds: Vec<String>`; one new predicate in `list_submission_instances_page` reads `data->'data'->'execution_policy'->>'task_kind'` (COALESCE to `'prompt'`, matching the Rust default) so fine kind discrimination happens in SQL alongside the existing `task_statuses` pushdown.
- The handler maps `filter.kinds ∩ {Prompt, Review, Planner}` into the new field (`TaskKind::as_str` added, mirroring the serde snake_case representation).
- Each row is now projected exactly once: the single projection feeds the kind check, dedupe key, page cursor, and summary construction.
- Fail-closed behavior preserved: `runtime_submission_task_kind` still parses/validates every fetched row, so a malformed `execution_policy` still aborts the listing exactly as before.

## Tests

- New store test `submission_filter_discriminates_prompt_family_kinds_in_sql`: seeds planner/review/default-prompt instances, asserts `prompt_task_kinds=["planner"]` returns only the planner row and an empty list keeps the whole family visible.

## Verification

Fresh local runs against throwaway PostgreSQL 16:

- `cargo test -p harness-workflow --lib` — 748 passed / 0 failed
- targeted: `submission_filter` filter 2 passed
- `runtime_submission` server filter — 77 passed
- `cargo test -p harness-server --lib` — **1626 passed / 0 failed**
- `cargo fmt --all -- --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean


majiayu000:
Pushed 93381e0d addressing all three open review findings:

- **GitHub rows dropped from mixed queries** (Gemini critical / Codex P2): the prompt-kind predicate is now scoped by `definition_id <> 'github_issue_pr'`, matching the Rust-side classification order.
- **Corrupt policies vanish under filters** (Codex P2): rows whose persisted `execution_policy.task_kind` falls outside `TaskKind`'s serde values are always kept in the fetched set (`<> ALL()`), so `runtime_submission_task_kind` still rejects them fail-closed instead of listings silently hiding them.
- **Duplicate `as_str`** removed in favor of the existing `AsRef<str>` impl.

New store test seeds a github_issue_pr row plus a corrupt-policy row alongside the prompt family and pins both behaviors.

majiayu000:
Pushed 8013cce6: the CI failure (`submission_filter_discriminates_prompt_family_kinds_in_sql`) was a real SQL bug in my keep-arm — `x <> ALL('{}')` is vacuously TRUE in PostgreSQL, so an empty `recognized_task_kinds` set leaked every row past strict kind filters. The arm is now guarded on `cardinality($17) > 0`; callers that never opt in keep pre-existing semantics. Verified against a disposable local PostgreSQL: full harness-workflow lib suite 749 passed / 0 failed, including both the CI-failing test and the new corrupt-policy regression test.

PR #2034, opened 2026-09-02
refactor(cli): decouple reconcile from server

### LINKED ISSUE #1953: refactor(cli): harness-cli links the entire server stack (fat binary)
## Problem

`crates/harness-cli/Cargo.toml` depends on all 12 workspace crates, including `harness-server` (and transitively `db-postgres` on core, the OTel stack, sqlx, etc.). `commands/serve.rs` is a thin wrapper into `harness_server`.

Consequences:

- Every CLI invocation (`harness gc`, `harness rule`, `harness skill`, `harness plan`) compiles and links the full HTTP/workflow/Postgres/OTel stack.
- CLI and server cannot be versioned, shipped, or hardened independently.
- CLI build times and binary size track the largest crate in the workspace (harness-server, ~134k LOC).

## Suggested direction

- Split `serve` into its own binary (or feature-gate the server dependency) so the day-to-day CLI paths only need core/protocol/workflow-client surface.
- Longer term, have CLI talk to a running server over HTTP/JSON-RPC via `harness-protocol` instead of linking server internals (relates to the intent of #1800).

Found during an architecture review of origin/main (9634e5a9).

majiayu000:
## Summary

- move the reconcile query and report wire contract into `harness-protocol`
- use protocol-enforced REST adapters in the server route
- compile and run the HTTP reconcile client without the `server` feature, leaving `serve` as the only in-process server command

## Validation

- `cargo check -p harness-cli --no-default-features --all-targets`
- `cargo test -p harness-server reconciliation_payload`
- `cargo check --workspace --all-targets`
- network-independent reconcile CLI tests passed individually

The aggregate `cargo test -p harness-cli --no-default-features commands::reconcile` command compiled successfully, but this runtime denies loopback TCP connections (`Operation not permitted`), so its mock HTTP tests could not complete locally.

Closes #1953


majiayu000:
Independent review (did not author this PR).

Reconcile query/report move onto `harness-protocol` RestDto, and the CLI talks HTTP without the `server` feature. Server route uses the contract adapters. This is a bounded slice of #1953, not the full fat-CLI close.

No blockers.

APPROVED

Issue #1977, opened 2026-08-22
workflow: WorkflowRuntimeStore has grown to 170 pub methods across 38 files


majiayu000:
## Execution disposition (2026-09-05)

Deferred after supervised runtime assessment. Method/file counts are maintenance signals, not sufficient acceptance criteria for a new role-trait layer. Resume only with a concrete caller boundary or testing problem and a bounded change; do not introduce a family of traits solely to reduce the apparent interface size. Recount current source before reusing the historical numbers below.

## Problem

`WorkflowRuntimeStore` now exposes 170 `pub fn`/`pub async fn` methods spread across 38 files under `crates/harness-workflow/src/runtime/store/` (up from ~133 methods across fewer files at the 2026-07-25 audit). Individual files are reasonably sized, but the *type* still owns job-claim, memory-extraction, remote-facts, decisions, evidence, retention, dead-letter handling, and PR-binding-repair as one undifferentiated interface.

The newer eval subsystem (~10.3k lines) and stack-capability subsystem (~7.8k lines) both now reach into `WorkflowRuntimeStore` directly rather than through any narrower contract, so the surface keeps growing as new features land.

## Why it matters

A god *interface* (even when physically split into small files) means every new subsystem's dependency is "the entire store," which makes it hard to reason about who can touch what, hard to mock/test in isolation, and increasingly expensive to split later as more callers accrete.

## Suggested direction

Segment `WorkflowRuntimeStore` into role-scoped traits (e.g. `JobLeaseStore`, `MemoryStore`, `EvidenceStore`, `RetentionStore`) that the concrete store implements, and have new subsystems (eval, stack-capability) depend on the narrow trait they actually need instead of the full store type.

Found during an architecture review of origin/main (0a3ebbb7).
## Maintainer disposition (2026-09-16)

Close as not planned. The maintainer accepted a backlog focused on delivering the reviewed runtime/shared-workflow changes and validating real task completion. Interface method/file counts do not establish a defect or justify role-trait extraction. A future concrete caller-boundary or testing defect should be tracked as a new bounded issue with reproduction and acceptance criteria. This closure does not claim the historical refactor was implemented.

Issue #1958, opened 2026-08-13
refactor(server): converge the dual task lifecycle models (legacy task_runner vs workflow runtime)


majiayu000:
## Current scope (2026-09-05)

The historical duplicate-enum claim is no longer accurate: task_runner/mod.rs re-exports TaskKind, TaskStatus, TaskState, and related types from workflow_runtime_submission. TaskStore and runtime-to-task projections still exist; type reuse does not complete lifecycle convergence.

Prioritize concrete incorrect state, timestamp, or persistence behavior over a full rewrite. PR #2041 addresses the observed instance updated_at bug. The current blocked-to-waiting projection is explicit in runtime_projection/mod.rs and must be distinguished from a workflow stuck in implementing. Do not introduce a compatibility layer or combine this work with #1798/#1976 without a demonstrated requirement.

## Problem

Two parallel task lifecycle stacks coexist:

- Legacy: `harness-server/src/task_runner/` (~3k LOC) with `TaskStore` and its own `TaskStatus`
- Workflow runtime: `workflow_runtime_submission/runtime_models.rs` defines another `TaskKind` / `TaskStatus` / `TaskState`

The crate description itself lists both "task runner" and "workflow runtime". Reconciliation and dashboards bridge the two lifecycles instead of one model owning the truth (e.g. `reconciliation_tests.rs` opens both stores).

## Why it matters

Operators, APIs, and metrics must reason about two state machines for the same concept; every cross-cutting feature (recovery, budgets, monitoring) pays the seam tax twice, and status-mapping bugs hide in the bridge code.

## Suggested direction

Decide the target model (presumably the workflow runtime), route new work through it only, and shrink `task_runner` to a compatibility shim with an explicit removal plan. Relates to #1798 (RuntimeContext extraction) — converging the models first makes that extraction smaller.

Found during an architecture review of origin/main (9634e5a9).

PR #2094, opened 2026-09-19
fix(agents): report selected model in Codex execute and review responses

majiayu000:
## Summary

Replace the hard-coded backend label in Codex oneshot response metadata with the model selected for the actual CLI launch. Add regression tests that invoke a local shell stub, not a provider or paid model.

Base reviewed: `b1be2763f7760e884b4640287a346c7d60fce824`.

## Problem and evidence

Both `CodexAgent::execute` and `CodexAgent::execute_review` pass a selected model using `-m`, but return `AgentResponse { model: "codex", ... }`. The backend label discards both configured defaults and per-request overrides. The streaming path already emits `ModelReported` with `ModelIdentitySource::LaunchDerived`.

This is a metadata defect. It does not establish incorrect provider billing, and this PR does not claim provider-confirmed serving-model identity.

## Design

- `execute`: populate the response from the existing `launch_model(&req)` selection used for `base_args`.
- `execute_review`: use the same request-over-default precedence already used by `review_args`.
- Keep `name() == "codex"` for backend identity.
- Keep the existing `AgentResponse` schema. Do not add a parallel model identity framework, compatibility shim, pricing table, or historical backfill.
- Treat this value explicitly as launch-derived, not provider-reported. The existing streaming source tag is unchanged.
- Leave cancellation, cleanup, protocol resource bounds, and runtime snapshot behavior out of this small PR; those require separate lifecycle/persistence designs and failure-injection tests.

## Regression coverage added

`codex_model_response_tests.rs` contains two Unix tests, one for execute and one for review. Each covers:

1. No request override: configured default is passed and returned.
2. Request model A: override is passed and returned.
3. Request model B: a distinct override remains distinct.

The stub reads the actual `-m` argument and echoes it through the normal output path, so assertions cover launch/response consistency rather than only a helper. Tests also assert successful exit and unchanged backend identity. Every execution is bounded by a timeout. No real Codex binary or credentials are required.

## Validation

- Fresh independent review and main-integration review approved synchronized head `f8db610c4b0aaa5d1a7c864e9d6dc4d418d14489`.
- [Latest-head CI run](https://github.com/majiayu000/harness/actions/runs/35462673841) passed Format, workspace/all-targets Clippy, Test, and CI Result.
- The Test job runs `cargo test --workspace --exclude harness-server`, including the complete harness-agents library suite and both new model-response regression tests. Package-level evidence comes from CI.
- No local Rust commands are claimed as run for this metadata change. No live provider or paid-model trial was performed.
- After main synchronization, the patch preserves the reviewed change. The earlier integration changed only index hashes and hunk line numbers after the separate prompt-terminator fix; independent integration review confirmed no model-selection interaction.

The reported identity remains launch-derived and does not prove the provider's serving model. Lifecycle and runtime-store follow-ups remain in #2095 and #2096.


majiayu000:
## Follow-up design and validation update

Latest submitted head: `e14809939f8b4c07c78508da02b1643ec53a88f6`.

The initial CI run identified three rustfmt-only differences. Applied the exact formatter output in commits `154a25debee3f312c778c7dc73b51efe1d00fa61` and `e14809939f8b4c07c78508da02b1643ec53a88f6`; inspected the resulting source diff. No behavior or scope was added.

### CI evidence observed

- Initial head `29f861e...`: workspace/all-targets Clippy passed, and the `cargo test --workspace --exclude harness-server` step passed. This is previous-head evidence, not a claim that all latest-head checks have passed.
- Latest head `e148099...`, [CI run 35417290832](https://github.com/majiayu000/harness/actions/runs/35417290832): Format passed, Web Build passed, Storage Legacy Openers passed. Test and Clippy were still running at the last check.
- Latest-head PR Check, Harness PR Pipeline, and Disabled Modules Check completed successfully.
- No local Rust commands were run. No independent fresh-context review has been claimed. Keep draft until latest-head `CI Result` and independent review are complete. No auto-merge is enabled.

### Remaining review items have full implementation designs

- #2095: cleanup errors must propagate; attempt-scoped startup cancellation; bounded stop progress and handshake/write deadlines; stale-generation protection; bounded protocol input. Includes scoped PR sequence, failure schedules, deterministic test matrix and merge criteria. The lifecycle changes are not implemented in this metadata PR.
- #2096: narrow runtime store-key hardening and regression plan, with explicit correction of the earlier snapshot concern. `AppState::persist_runtime_state` already locks before snapshot capture and retains the lock through persistence, so the same-AppState out-of-order-write concern is withdrawn. Normal startup also creates the directory first. Do not introduce a CAS/schema redesign or claim a proven normal-startup data-loss bug.

Existing lifecycle convergence stays with #1958; this review does not justify resurrecting the broad Store-trait split deferred in #1977.

